PR #49, opened 2015-10-07
Compile on Node 4.x

mbilker:
I was able to make a minimal amount of changes to create a successful build using Node 4.1.2. Though this build breaks backwards compatibility with the Electron upgrade to 0.30.7.


gaborbernat:
builds and runs on latest arch :+1:

emorikawa:
@mbilker thank you so much for taking the effort to track all of this down. We actually were just looking at this issue ourselves, but will definitely use your pull request! The one big thing in here is the Electron bump. We (and the Atom team frankly) have unfortunately been trained to be very wary of Electron version bumps for the sake of stability. I really appreciate you doing the research and picking the 0.30 line (before the Chromium bump). I was just curious if there was a particular thing in the 0.30 line that you needed to get this working? Regardless, we also think the 0.30 upgrade is good (and in fact may do 0.30.8 instead of 0.30.7).

mbilker:
Wow. I used 0.30.7 because it is used in the current stable version of
Atom. I compared the dependencies of the package.json in the build folder
and noticed the significant changes for the packages regarding the grunt
electron usage and how outdated they were. I then set the versions to
Atom's corresponding version. After many test builds, the build worked and
every spec test passed.

TL;DR Used Atom's build system changes to make N1 build under node v4

dillonforrest:
Hey there @mbilker, thanks for your time in submitting this pull request!

I'm really sorry to do this, but can I trouble you to sign our CLA? https://nylas.com/cla

I know it might be kind of annoying, but our stakeholders require all our contributors to sign our CLA. We made ours similar to [Facebook's CLA](https://code.facebook.com/cla). It'd mean a lot to us if you could sign it! :)

mbilker:
@dillonforrest I signed the CLA. It's under Matt Bilker <me@mbilker.us>.

PR #50, opened 2015-10-07
fix(docs): correct SMPT->SMTP

lukegb:
Looks like a small typo slipped through :)


emorikawa:
Yes! Great catch. Thank you so much for the attention to detail! Also, thank you so much for signing the CLA for this little patch. I know it's a pain for stuff like this.

PR #52, opened 2015-10-08
Change atom to nylas in clean script

mbilker:
I discovered in the clean script that there are some remnants of the Atom editor project. I refactored the names to their current directory names.


bengotow:
This looks great—thanks for fixing @mbilker.

PR #53, opened 2015-10-08
Travis update

mbilker:
I noticed on the official N1 travis page it succeeded, but the actual spec tester failed run because it couldn't connect to a X11 display. I took the liberty of revamping the travis build script to follow Atom's example and build separately on Linux and OS X.

The Linux build is still having issues, but I do not know how they are occurring. I probably will setup a Ubuntu Precise box just to solve the issues. The OS X build completes spectacularly with the spec test being run, and all of them passing.

EDIT: The Travis script uses a developer Github API OAuth key. Just follow the instructions at https://developer.github.com/v3/oauth/ to get a access token. Or continue using the one I generated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,16 @@
-language: node_js
-sudo: true
-before_install:
-  - sudo apt-get install libgnome-keyring-dev
-install:
-  - ./script/bootstrap
-env:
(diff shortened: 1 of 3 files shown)


emorikawa:
@mbilker This is fantastic. Thank you for diving into this.

Issue #28, opened 2015-10-06
Windows build


iamchathu:
I would like to try N1 in Windows 8.1 (64-bit). I installed cloned the repo and installed the dependencies.How to build the repo for windows.
I tried running build.cmd i get an error

```
Node: v0.12.7
npm: v2.11.3
Python: v2.7.9
F:\GitHub\N1\script\bootstrap:44
    throw new Error("Building for win32 with architecture "+targetArch+". Are
          ^
Error: Building for win32 with architecture x64. Are you sure you meant to do th
is?
    at makeSqlite3Command (F:\GitHub\N1\script\bootstrap:44:11)
    at bootstrap (F:\GitHub\N1\script\bootstrap:72:31)
    at F:\GitHub\N1\script\bootstrap:192:3
    at F:\GitHub\N1\script\utils\verify-requirements.js:21:9
    at F:\GitHub\N1\script\utils\verify-requirements.js:115:5
    at ChildProcess.exithandler (child_process.js:742:7)
    at ChildProcess.emit (events.js:110:17)
    at maybeClose (child_process.js:1015:16)
    at Process.ChildProcess._handle.onexit (child_process.js:1087:5)

```

gardner:
Try with node version 0.10.28

gardner:
Your's looks like a different problem but currently N1 is only building against node v0.10.x. Maybe give that a try?

jsjan:
``Error: Building for win32 with architecture x64. Are you sure you meant to do th
is?``
should explain this. You said you run Win8.1 with 64-bit (which is x64), somebody should try this on a 32-bit Windows.

iamchathu:
I'm using Win8.1 64bt.I installed dependancies and then ran build.cmd in scripts folder.

slnj:
@iamchathu Thats not the point, I mean Win32 is the 32-bit API for Windows so it should work with a 32-bit Windows.

bengotow:
Hi folks—I'm the author of that not-very-helpful error in script/bootstrap.

Earlier this year, there was only a 32-bit version of Electron for Windows. We put in that check to ensure that our CI server never got moved to a 64-bit box, since it would compile sqlite in 64-bit and we'd be loading it into a 32-bit version of NodeJS.

This is no longer the case — Electron became available for 64-bit Windows in April (https://github.com/atom/electron/issues/980). Would love to know if you're able to remove that warning and build the whole thing for 64-bit Windows.

slnj:
[...]

PR #68, opened 2015-10-09
fix(script): allow x64 build on Windows

jan509:
Electron is available for 64-bit Windows machines(https://github.com/atom/electron/issues/980), so we don't need the 32-bit check anymore. Reference: https://github.com/nylas/N1/issues/28

@bengotow Is this ok for now?

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -39,10 +39,6 @@ function makeSqlite3Command() {
   var targetPlatform = require('os').platform();
   var targetArch = require('os').arch();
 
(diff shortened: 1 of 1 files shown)


bengotow:
Hi @solnj — excited we can remove that warning now! If you've tested the build on a 64-bit Windows machine and everything is working, I think we can safely delete it. 

Could you take a moment and sign our open-source contributors CLA? https://nylas.com/cla. Then I can land this—thanks for contributing to the project!

solnj:
@bengotow Signed!

emorikawa:
@jan509 Thank you!

PR #70, opened 2015-10-09
Updated comments, cleaner to read

zachpanz88:
Made some changes to the format of comments that makes them easier to read, did not change any actual content of the comments.

### DIFF
--- a/static/buttons.less
+++ b/static/buttons.less
@@ -18,9 +18,9 @@ button, html input[type="button"] {
   img.content-mask { background-color:@btn-default-text-color; }
   background: @background-gradient;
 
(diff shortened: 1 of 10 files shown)


mbilker:
How is this easier to read?

zachpanz88:
It may not exactly be easier to read, but it provides an overall cleaner look

grinich:
Thanks for checking out N1, @zachpanz88. Our style choices are to preface even multi-line comments with the `//` mark, so we unfortunately won't be merging this.

We'd love to get your thoughts+edits on the other functional parts of N1, however! :smile:

PR #88, opened 2015-10-13
improve readme / rewrite "open-source" to "free software"

slang800:
While this program is obviously "open source", merely having the source available [misses the point](http://www.gnu.org/philosophy/open-source-misses-the-point.html) of the GPL. It's much more important that the program respects the users' freedoms, and the fact that it does should be praised in the readme.

I also moved the link to the Nylas Sync Engine on to its name, which I think makes more sense since the link is going to the main page for the project and added a link to define "free", just in case anyone mistook it for meaning "no cost".


mbilker:
Somewhat related to this, the project is missing a copy of the GPLv3. It references the GPLv3 in the LICENSE file, but it also says a copy of the GPLv3 should be included with the source code of this program.

slang800:
@mbilker - good point. I'll add that into the PR

mbilker:
Your change broke the generate-license grunt task. Also, I believe you should keep the LICENSE.md file, it still explains who the project's copyright belongs to.

mbilker:
A recent PR accomplished partly the same thing as this.

slang800:
@mbilker - if you're talking about #453, I saw that and fixed the changes it introduced in #491... Otherwise I'm not sure which PR you're talking about.

As for the grunt task - fixed that and re-pushed... right now the build seems to be failing due to some gyp issue that's specific to node 0.12 on linux.

mbilker:
@slang800 ahh. It failed to download Electron. Also, please readd LICENSE.md. The grunt task `generate-licenses` uses that file.

slang800:
I [switched](https://github.com/slang800/N1/blob/5552fa5ec97a8b08fc858ee238134789f62e4791/build/tasks/generate-license-task.coffee#L29) the grunt task to use the new `LICENSE` file, since the GPL (sadly) wasn't written in markdown.

PR #113, opened 2015-10-21
Setup stylesheets for dropdown (select) control

agudulin:
Default select tag looks creepy, so here are some changes that make dropdowns look way better.


bengotow:
Hi @agudulin — this looks great. We may refine the style of these a bit more in the future (particularly on linux / windows) but this is definitely an improvement over the system standard ones. Thanks!

PR #114, opened 2015-10-21
Rename some variables from atom to nylas

mbilker:
The commits in relation to this pull request change the variable names from `atom` to `nylas` in the build scripts. I do not know if they should be `nylas` or `n1`.

Only commit 3a5e0f0 renames `atom.coffee` to `nylas.coffee`. I found the referenced files associated with the renamed file and altered them to point to the new file.

The name for a563f1d is not what the title says. I changed `engines?.atom?` to `engines?.nylas?`. That commit reverts that. Also, there were many other locations I would need to change and test if regressions occur.

### DIFF
--- a/build/Gruntfile.coffee
+++ b/build/Gruntfile.coffee
@@ -4,8 +4,8 @@ os = require 'os'
 
 # This is the main Gruntfile that manages building N1 distributions.
 # The reason it's inisde of the build/ folder is so everything can be
-# compiled against Node's v8 headers instead of Atom's v8 headers. All
-# packages in the root-level node_modules are compiled against Atom's v8
+# compiled against Node's v8 headers instead of Electron's v8 headers. All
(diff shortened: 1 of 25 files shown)


bengotow:
Thanks for submitting this PR! This is something we've been meaning to clean up for a while. I'll review these changes this week and make sure the CI builds still run properly on mac/win/linux before we land it. Stay tuned.

mbilker:
@bengotow Thank you very much! I haven't fully tested the build process on Windows (as I don't use Windows on a regular basis), but I do use N1 on my Mac and Linux machines.

mbilker:
The CI is still not working on 0.12 which is strange.

mbilker:
#51602f69a54ee11b72e624ba7e63503f329a75d1 implements the changes in this PR, but better.

PR #203, opened 2015-10-26
Update README.md

Stoner19:
Instructions for users to compile N1 themselves added to README.md


Stoner19:
I know this information is included in the CONTRIBUTING.md doc, but not many would look there for compiling instructions.

bengotow:
Hey! Thanks for submitting this—the `CONTRIBUTING.md` file gives a lot more context, and also has instructions for running the sync engine, so I think we should update the README.md to point to that rather than duplicating it into the readme.

Stoner19:
I think that's a great idea. feel free to close this pull request, and just add a link to your CONTRIBUTING.md doc.

PR #240, opened 2015-10-28
Fixed building rpms

Zeklandia:
Many file paths had had `/usr/local/share/nylas` instead of `/usr/local/share/Nylas`
I created `publish-build-task.coffee` as an attempt at fixing an error, it might not be necessary.


bengotow:
Hi @Zeklandia — thanks for submitting this pull request. I'm not sure if we should convert everything to uppercase (`usr/local/share/Nylas`) or lowercase. I think other things in the share directory are typically lowercase.

Zeklandia:
Right @bengotow, they just have to all be the same. :)

bengotow:
Hey @Zeklandia —I squashed these and reverted one tiny change to a Nylas.exe binary, and landed it in 8cbcace3ef1025ba22a84f653d2ff060972780a2.

PR #301, opened 2015-10-30
Update debian package libgcrypt dependency

nicholascapo:
* Package libgcrypt11 is no longer available in Debian or Ubuntu
* Allow installing the package with libgcrypt20 instead

I was not able to install the Official .deb without this modification.

But after repacking the .deb with this change included, I was able to install and run N1 with no problems.

Debian Package Search:
https://packages.debian.org/search?keywords=libgcrypt&searchon=names&suite=stable&section=all

Ubuntu Package Search:
http://packages.ubuntu.com/search?keywords=libgcrypt&searchon=names&suite=wily&section=all


spang:
Thanks for your contribution @nicholascapo! We'll make sure this works on our CI server and then hopefully folks will stop having this problem. :)

PR #373, opened 2015-11-06
Travis CI - Add support for Node 4.2 LTS

mbilker:
This pull request adds support for node 4.2 for both Linux and OS X. The build environment for Linux at Travis uses Ubuntu 12.04 which includes GCC 4.6 that doesn't support C++11 extensions required by NAN for building under Node 4.2.

I added the `ubuntu-toolchain-r-test` repository to install a newer GCC 4.8 that can build Nylas N1 under Node 4.2.

For some strange reason, some builds on OS X under Node 0.12 fail during the NPM download stage with Connection Reset errors.

EDIT: Change 4.1 to 4.2. Node v4.2 is the LTS release. Node v5 is the latest/stable release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,14 @@ matrix:
       env: NODE_VERSION=0.10
     - os: linux
       env: NODE_VERSION=0.12
-    - os: linux
-      env: NODE_VERSION=4.1
+# Wants a c++11 compiler
+#    - os: linux
+#      env: NODE_VERSION=4.1
     - os: osx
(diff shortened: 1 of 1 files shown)


mbilker:
The failure was caused by the same error that randomly occurs. I think there is a race condition in `ThemeManager`.

bengotow:
Hey! I've merged this as 87bc41f600fc2351c6615613affb4cf982b14f86 — thanks for fixing this, I was wondering how we were going to fix the `nan` issues!

PR #396, opened 2015-11-11
Allows package to be installed in ElementaryOS

luixal:
Changes the condition for detecting the distro name so ElementaryOS (ubuntu based) goes as an Ubuntu and the package installs.

### DIFF
--- a/build/resources/linux/debian/postinst
+++ b/build/resources/linux/debian/postinst
@@ -24,7 +24,7 @@ case "$1" in
 
         DISTRO=`lsb_release -s -i`
 
-        if [ "$DISTRO" = "Ubuntu" ]; then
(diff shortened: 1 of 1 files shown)


bengotow:
Hi @luixal — sorry for the delay, this is great. Can you sign the Nylas CLA at https://nylas.com/cla? That'll make the PR bot happy and we can land this :-) Do you know if there are other distro names that should be listed there? I imagine ElementaryOS isn't the only exception, but we can definitely start with it!

luixal:
Hi @bengotow -- I've just signed the CLA, so... I gues everything is good to go. Not really sure about other distros, but all debian/ubuntu based ones (Linux Mint for example) should go the same way, but I haven't tested the patch on them, just on eOS.

To find more ubuntu based distros, you can have a look here: [Ubuntu based distros list on distrowatch ](http://distrowatch.com/search.php?basedon=Ubuntu).

Issue #412, opened 2015-11-17
FileDownloadStore removes fulfilled download in _cleanupDownload before trigger


mbilker:
In `src/flux/stores/FileDownloadStore.coffee`, I noticed in `_cleanupDownload` on line 233-236 that you trigger the store update callbacks after deleting the fulfilled event.

This prevents me from using the event to detect when an attachment download finished in my PGP plugin. The events completed events are not present in the `Object` so I cannot detect the change.

![2015-11-16-223448_1280x800_scrot](https://cloud.githubusercontent.com/assets/1266541/11202291/9da5ec8e-8cb4-11e5-8def-a6c873c61df7.png)

mbilker:
Don't worry about the double logging, I placed 2 `console.log`'s in different places and forgot about one.

mbilker:
Created a pull request, Travis running now.

bengotow:
Hey! I think this has been addressed with your PR, gonna close this.

mbilker:
Ahh yes it has. Thank you!

PR #413, opened 2015-11-17
fix(FileDownloadStore): delete download from @_downloads after @trigger

mbilker:
From #412:

In src/flux/stores/FileDownloadStore.coffee, I noticed in _cleanupDownload on line 233-236 that you trigger the store update callbacks after deleting the fulfilled event.

This prevents me from using the event to detect when an attachment download finished in my PGP plugin. The events completed events are not present in the Object so I cannot detect the change.


mbilker:
I already signed the CLA.

bengotow:
Thanks @mbilker — this looks good! I'll make sure it goes into the next beta.

PR #441, opened 2015-11-21
Remove vm-compatibility-layer

mbilker:
`vm-compatibility-layer` was designed to support obsolete APIs in node v0.11.x of node v0.10.x. It is only used in the build Gruntfile and nowhere else in the project.

I ran the build without `vm-compatibility-layer` and it completed with no errors, thus demonstrating its nonessential.

Atom already removed `vm-compatibility-layer` a few versions ago.


bengotow:
Hi @mbilker — this looks good to me. It looks like the build scripts still work with Node 0.10 and Node 0.12. If anyone runs into issues running the build scripts on another version of Node we can put this back as a build dependency only.

Issue #436, opened 2015-11-20
Messages get same body as other message


mbilker:
Major bug for me. When testing my email-pgp plugin against multiple Facebook encrypted emails. I noticed it only decrypted the message for the first one, and used the cached version for the rest of the messages. It does not even hit my `MessageStoreExtension` to ask to format the other messages.

Here is an example:
![2015-11-19-224152_1280x800_scrot modified](https://cloud.githubusercontent.com/assets/1266541/11291558/d30be362-8f0f-11e5-9183-e3ce6f588df6.png)
vs.
![2015-11-19-224702_1280x800_scrot modified](https://cloud.githubusercontent.com/assets/1266541/11291551/c14ad6ec-8f0f-11e5-936f-ceafe41a3926.png)
which should be
![2015-11-19-230538_1280x800_scrot modified](https://cloud.githubusercontent.com/assets/1266541/11291772/6bc62e8a-8f12-11e5-9065-a39f566a023a.png)

mbilker:
The issue is here: https://github.com/nylas/N1/blob/master/src/flux/stores/message-body-processor.coffee#L40-L41

It generates the key for the message before processing the message through the extensions. The body on the Facebook encrypted emails is `""` so the key for all the Facebook emails is the same and it doesn't reprocess the message.

mbilker:
In the Nylas Developer Tools console, the MD5 hash of `""` results in
```
require('crypto').createHash('md5').update("").digest('hex')
"d41d8cd98f00b204e9800998ecf8427e"
```

mbilker:
I don't know how to workaround the issue for the N1 project as a whole. Those message extensions can be computationally expensive.

I understand hashing the original body and caching the output of the extensions with the hash of the original body speeds up the process dramatically.

mbilker:
I accidentally removed a block of code in mbilker/email-pgp#ba8c540 and readded it in mbilker/email-pgp#1a1048d

bengotow:
Hi @mbilker — hmm, interesting... so the message body is identical in those messages, but the output /after/ the message body extensions run is unique?

If that's the case, I think we should change that hash function to use more than just the body to generate the hash. For example, we could say:

```
require('crypto').createHash('md5').update(message.id + message.body + message.subject).digest('hex')
```

Overall, the goal of that caching is to avoid re-processing the same message twice as messages are collapsed / uncollapsed, etc. and React renders are being triggered. It's alright if it runs for each message once.

[...]

PR #443, opened 2015-11-22
Do not return when original body is null

mbilker:
39a142ddcb80c7e1fce22dfe1e0e628272154523 fixed the prime issue of #436, but introduced another issue. Facebook emails have a `""` body initially. The return statement in the `process` method of `MessageBodyProcessor` causes the method to return early, ignoring my extension that would set the body at https://github.com/mbilker/email-pgp/blob/master/lib/message-loader/message-loader-extension.es6#L19. This PR removes the return statement, but leaves in the change in the hash key for caching messages.

### DIFF
--- a/src/flux/stores/message-body-processor.coffee
+++ b/src/flux/stores/message-body-processor.coffee
@@ -1,3 +1,4 @@
+_ = require "underscore"
 crypto = require "crypto"
 MessageUtils = require '../models/message-utils'
 MessageStore = require './message-store'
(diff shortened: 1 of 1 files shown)


bengotow:
Hi @mbilker — I think we want the process method to return when the body is `null`, implying that it hasn't been downloaded yet, but not when the body is `""`. I keep forgetting that `"" == false` in JavaScript. I think this line should be `return "" unless _.isString(body)` to be more clear and fix the issue you're running into.

The message body processing breaks lower down if the body is `null` or `undefind`, so we need to check for those cases before proceeding.

mbilker:
Hey @bengotow, I now agree with you. The processing does break down if the body is `null` or `undefined`. When you push the change you (or I) can close this pull.

PR #453, opened 2015-11-24
Internal package.json license: GPLv3

dcposch:
Changed from 'Propietary' to 'GPLv3' to match `LICENSE.md`

### DIFF
--- a/internal_packages/account-sidebar/package.json
+++ b/internal_packages/account-sidebar/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "main": "./lib/main",
   "description": "Sidebar view that shows your account and important tags",
-  "license": "Proprietary",
+  "license": "GPLv3",
   "private": true,
   "engines": {
     "nylas": "*"
--- a/internal_packages/attachments/package.json
+++ b/internal_packages/attachments/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "main": "./lib/main",
   "description": "View attachments on messages",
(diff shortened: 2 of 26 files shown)


dcposch:
Alternatively we could just remove the `license` fields from each `package.json`. N1 says it's GPL in documentation & LICENSE.md, don't know if we have to specify again for each individual internal package. Thoughts?

bengotow:
Hey @dcposch — thanks for pointing this out, we definitely need to fix this. Can you do me a favor and sign the Nylas CLA (https://nylas.com/cla)? That'll make the PR bot happy and we can land this.

It's not really necessary to have the license in there since the entire project is GPLV3, but I think these packages may be copy-pasted a lot as people create their own, and we'd like for all of those to list a license, so I vote we keep the field.

dcposch:
@bengotow done. Should I also change the example packages, eg https://github.com/nylas/N1/blob/master/examples/N1-Filters/package.json ?

internal_packages/account-sidebar/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "main": "./lib/main",
   "description": "Sidebar view that shows your account and important tags",
-  "license": "Proprietary",
+  "license": "GPLv3",
```
slang800: the [SPDX license code](https://spdx.org/licenses/) is `GPL-3.0`... npm won't recognize `GPLv3` properly

slang800: update: fixed that in #491

PR #525, opened 2015-12-01
Change backspace behavior

Yerlix:
There is a setting so that the backspace can delete messages instead of archiving.
I think that backspace must delete the message, but like many other clients, you have the choice.


internal_packages/thread-list/lib/thread-list.cjsx
```diff
@@ -284,13 +284,18 @@ class ThreadList extends React.Component
...
+        if FocusedMailViewStore.mailView().canTrashThreads()
+          removeMethod = TaskFactory.taskForMovingToTrash
+        else
+          return
       else
-        return
+        if FocusedMailViewStore.mailView().canArchiveThreads()
+          removeMethod = TaskFactory.taskForArchiving
+        else
+          return
```
bengotow: Can we fall back to `removeMethod = TaskFactory.taskForMovingToTrash` on this line? I think that since backspaceDelete is false by default, users who don't have an archive folder will reach here, `canArchiveThreads` will return false, and nothing will happen when they press "delete" :-)

bengotow:
This is looking great! Can you sign the Nylas contributor CLA? (https://nylas.com/cla) That'll make the GitHub bot happy and I can land this :-)

Issue #411, opened 2015-11-16
Cannot locate electron from /usr/bin/nylas. /usr/share/nylas/nylas works fine


jeffeb3:
Installed using the provided deb lubuntu 14.04:

sudo dpkg -i N1.deb

    >which nylas
    /usr/bin/nylas

    >nylas
    Cannot locate electron. Be sure you have run script/bootstrap first from /usr/bin
    *sidenote* not a super helpful message...

    >/usr/share/nylas/nylas
    # runs seemingly normally.

I think you probably meant to have /usr/bin/nylas be a soft link to /usr/share/nylas/nylas (seems nicer to my bin dir, to say the least). Because it's being ran from /usr/bin, it's not seeing the things it sees in /usr/share/nylas. Just a guess.

mbilker:
This is because `/usr/bin/nylas` is the `N1.sh` script found in the base directory of the project (https://github.com/nylas/N1/blob/master/N1.sh). This script thinks the N1 project directory is `/usr/bin`, but since it cannot open Electron from `/usr/bin` it spits out the error (https://github.com/nylas/N1/blob/master/N1.sh#L63). Line 63 shows the error and `ELECTRON_PATH` is equal to `$N1_PATH/electron` and `N1_PATH=/usr/bin` because of `BASH_SOURCE`.

mbilker:
I do believe /usr/bin/nylas should behave more like Atom's `atom.sh` script at `/usr/bin/atom` where it checks if there is an installed version and a built version in `/tmp` (different on other OS's).

jeffeb3:
The atom script does seem to do the trick, because it knows where the actual atom, and electron are located. You could make your script smart enough to know that. But couldn't you just put the script in the location of the install, and then make the /usr/bin a link to that script? The script would always be in the install directory, so it would know where to find everything. It seems like that would be more flexible, because you would only need to generate a symlink when doing the install, and at that point, you would know best where it was. Seems like it would easily extend to user space installs, or different os installs (like suse, some stuff is in /opt).

mbilker:
```
 ~  file /usr/bin/atom
/usr/bin/atom: Bourne-Again shell script, ASCII text executable
 ~  head /usr/bin/atom
#!/bin/bash

if [ "$(uname)" == 'Darwin' ]; then
  OS='Mac'
elif [ "$(expr substr $(uname -s) 1 5)" == 'Linux' ]; then
  OS='Linux'
[...]

PR #593, opened 2015-12-06
Convert Atom's launch script for N1

mbilker:
Addresses #411

The `N1.sh` launch script fails to account for running from `/usr/bin/nylas` or `/usr/local/bin/nylas` or `/opt/N1/bin/nylas` where it would throw an error saying it couldn't find the electron binary.

The Atom launch script works just fine and takes that into account. The installed launch script here is a refactored Atom launch script that properly accounts for running the `nylas` executable and locates the actual binary accordingly.

### DIFF
--- a/N1.sh
+++ b/N1.sh
@@ -3,14 +3,11 @@
 N1_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd );
 
 if [ "$(uname)" == 'Darwin' ]; then
-  OS='Mac'
   ELECTRON_PATH=${ELECTRON_PATH:-$N1_PATH/electron/Electron.app/Contents/MacOS/Electron}
 elif [ "$(expr substr $(uname -s) 1 5)" == 'Linux' ]; then
-  OS='Linux'
   ELECTRON_PATH=${ELECTRON_PATH:-$N1_PATH/electron/electron}
   mkdir -p "$HOME/.nylas"
 elif [ "$(expr substr $(uname -s) 1 10)" == 'MINGW32_NT' ]; then
-  OS='Cygwin'
   ELECTRON_PATH=${ELECTRON_PATH:-$N1_PATH/electron/electron.exe}
 else
   echo "Your platform ($(uname -a)) is not supported."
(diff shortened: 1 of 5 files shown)


mbilker:
I did not incorporate the AppVeyor development branch into this PR.

mbilker:
The current AppVeyor setup requires access to the project secrets in order to build.

whitelynx:
Any chance this could be rebased and merged? It'd be nice if an installed version of N1 actually worked out of the box...

mbilker:
@whitelynx N1 works out of the box. This PR just fixes the `/usr/bin/nylas` script. You can still launch N1 via `/usr/share/nylas/nylas`.

mbilker:
@bengotow When can this be merged? Some people complained about this on Slack, but it is a low priority issue.

bengotow:
Hey @mbilker — sorry for the delay on this one. We talked about this internally, and it'd be great if we could add the relevant Linux bits of `nylas.sh` (lines 89-116 I think?) to the existing `N1.sh` file in the root of the repository. I think there are a lot of things in that `nylas.sh` file pulled from `atom.sh` that we don't need anymore, and it might be confusing for the repo to have both N1.sh and nylas.sh.

I think we should also edit `install-task.coffee` to copy the file to both `N1.sh` and `nylas.sh` in the user's bin directory so that they can also do `N1` from the command line.

mbilker:
I see what you are saying @bengotow. The current `N1.sh` in the root of the repository allows N1 to be run from source using the `electron` command, and my script does not take this into account.

I do notice some parts of this `nylas.sh` are not needed anymore as they pertain to only Atom's launch script. I do believe the output handling parts are essential so I don't have to type `ELECTRON_ENABLE_LOGGING=1` every time.

mbilker:
Is there anything else I need to do to the script?

emorikawa:
Hey @mbilker This looks great. Thank you! Sorry for the delay on this one. I just rebased this on top of master and pushed.

mbilker:
Thanks @emorikawa !

PR #628, opened 2015-12-08
Replace function accessor with property accessor in Grunt's task-helper script

mbilker:
This bug appeared during a Travis test run, and only appeared on the Linux hosts running the tests. The function appears to treat the `process.env` as a set of functions, not properties.

This commit correctly interprets them as properties.

The main N1 repository no longer runs unit tests on Linux hosts, which is why this hasn't appeared during Travis builds, while my fork runs the unit tests on Linux under the Travis Ubuntu Trusty beta.

### DIFF
--- a/build/tasks/task-helpers.coffee
+++ b/build/tasks/task-helpers.coffee
@@ -77,9 +77,9 @@ module.exports = (grunt) ->
       false
 
   notifyAPI: (msg, callback) ->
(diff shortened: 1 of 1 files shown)


bengotow:
Thanks @mbilker — this looks great! I think we'll be turning the unit tests on linux back on soon. We were having a couple minor issues with our CI servers, but now that we're using Travis I think those should be taken care of.

PR #613, opened 2015-12-08
Travis build script rework to allow completion under Pull Requests

mbilker:
This PR stems from recent failures I encountered at [my forked repository](https://travis-ci.org/mbilker/N1) and with the PR's I sent here.

In nylas/N1@989f38ca77e56eaf3d8457b6c2ed03b14b5f6fcd the entire build script was merged from Evan Morikawa's branch for integrating AppVeyor, which reordered the whole Travis build script. This PR  reorders the to have the Travis build environment config at the beginning of the script and the build process at the end and the Travis secure variables at the end of the script.

I also added added some conditions to run the decryption commands only on the main repository using `TRAVIS_PULL_REQUEST` from [Testing Pull Requests](https://docs.travis-ci.com/user/pull-requests).

PR #629, opened 2015-12-09
Require `classNames` throws error under Linux

mbilker:
The issue stems that Mac OS X uses a case-insensitive filesystem by default, while most Linux filesystems are case-sensitive. The classNames package will require successfully on Mac OS X during Travis tests, but since the spec test suite isn't normally run under Linux this bug is more unlikely to appear.

This PR changes the module name to `classnames` instead of `classNames` to account for case-sensitive filesystems.

### DIFF
--- a/src/components/editable-list.jsx
+++ b/src/components/editable-list.jsx
@@ -1,5 +1,5 @@
 import _ from 'underscore';
-import classNames from 'classNames';
+import classNames from 'classnames';
 import React, {Component, PropTypes} from 'react';
 
 class EditableList extends Component {


bengotow:
Ahh good catch @mbilker! Not sure what's up with the CI, landing this.

mbilker:
Travis and AppVeyor do not provide the encryption keys to PR's for security reasons. PR #613 resolves that issue for Travis (not AppVeyor) by only running the decyption utilities for the main repository, not PR's.

PR #729, opened 2015-12-16
Display the folders name with a slash.

francois2metz:
When using imap folders, we don't want to display the sub folder separator like the point.

For *Archives.2008* it now display *Archives/2008*. It's also works when searching :mag: 

I copy/pasted the replace function from the [AccountSidebarStore](https://github.com/nylas/N1/blob/master/internal_packages/account-sidebar/lib/account-sidebar-store.coffee#L66). Do you know a place where we can reuse it?


emorikawa:
Hey @francois2metz. First of all I really appreciate the pull request. The code looks great. However, after talking about this more with the team, we decided that's it's probably best not to change the appearance of the label to users in the category picker. We actually use the presence of some of these characters to decide whether or not to nest folders on the account sidebar. I'm afraid one folder will be nested, and another not, but they both look like they have a slash in the label picker. The exact characters in the label is something that we should take care of (and fix if broken) on the sync-engine side and keep the display on N1 true to what we get from the API.

PR #792, opened 2015-12-21
Names with @ are included in the contact name.

### LINKED ISSUE #713: Sender Name Truncated if it contains the @ character
This is a minor issue with how Nylas handles sender names. 

It seems like if there is a `@` character in the sender name (such as `<NYL@S nylas@nylas.com>`) the sender name would only be recognized as NYL

image

In the message body: 
image

When expanded, it displays the full name. 

The full sender tag in this case was something along the lines of 
L@O 2012 <email@domain.com> 

This is on OSX with build 0.3.29-1904c6f 

I'm happy to poke around to try to see where the issue is. I suspect this is a regex pattern to handle cases of senders that do not provide a name field that greedily matches to the first `@`.

Sumukh:
This resolves #713. 

Tests have been added for this case but this change would parse a contact with the name "ev@nylas.com" to have a firstName of "Ev@nylas.com"

Previously this would have parsed to a firstName of "Ev". 

The only way around this is to introduce a email format check against names. I'm not sure how common it is to have a name field be a valid email address - but it might be more common than normal names that contain `@`.

### DIFF
--- a/spec/models/contact-spec.coffee
+++ b/spec/models/contact-spec.coffee
@@ -85,6 +85,45 @@ describe "Contact", ->
     expect(c3.displayFirstName()).toBe ""
     expect(c3.displayLastName()).toBe ""
 
+
+  it "properly parses names with @", ->
+    c1 = new Contact {name: "nyl@s"}
+    expect(c1.firstName()).toBe "Nyl@s"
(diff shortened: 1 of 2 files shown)


bengotow:
Looks good—thanks for expanding the test suite!

PR #798, opened 2015-12-22
AppVeyor Use no secure variables in Pull Requests

mbilker:
This PR wraps the statements requiring access to the secure variables in an `if` statement to check if the variables are set. This successfully builds N1, **but** I have no feasible way to check if N1 works under Windows. The build currently _fails_ with the `AWS_ACCESS_KEY_ID` environment variable not found for publishing the finished build to AWS S3.

The only computer I have Windows on is my old Lenovo X300 from 2008, which I really don't want to dust off and power it up.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,18 +28,18 @@ install:
 
 # We need to extract the encrypted private ssh key to clone the submodule.
 - ps: |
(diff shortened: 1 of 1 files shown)


bengotow:
Thanks - this looks good. Going to check that it works on master after landing.

mbilker:
Looks like it worked @bengotow !

Issue #794, opened 2015-12-21
Feature Request: Archive icon too similar to Trash icon.


colllin:
I find myself comparing them every time before I click one.

For reference, this is how Spark solves it:
![fullsizerender](https://cloud.githubusercontent.com/assets/185923/11935103/3008f7c8-a7c3-11e5-88c8-1a60da187948.jpg)

mbilker:
So you are suggesting to make the trash icon turn red?

colllin:
That seems to help. Also, their icons are slightly less similar to each other.

mbilker:
@colllin Alright then. What OS do you use for N1?

colllin:
OSX 10.11.2

mbilker:
Thank you!

mbilker:
@colllin So I did an small change. Here is with the background set:

![2015-12-21-22-45-20_359x57](https://cloud.githubusercontent.com/assets/1266541/11947445/d935dc46-a834-11e5-895d-8e2d5eb92de6.png)

Here is with the color changed:

![2015-12-21-22-50-43_379x74](https://cloud.githubusercontent.com/assets/1266541/11947472/46dfb140-a835-11e5-9904-4f6952321258.png)

colllin:
Nice! To me those are both more clear than it is now. Do you think so?

mbilker:
@colllin Yes. I believe the one with only the trash can red, not the background, is better if the color was a darker shade of red.

mbilker:
@colllin How do you think this looks?

![2015-12-21-23-00-46_339x79](https://cloud.githubusercontent.com/assets/1266541/11947555/b05aef08-a836-11e5-9b4b-c0d987bcd8f6.png)

colllin:
I think it's great. Are other people feeling the same way? It's also possible that I got used to the colored distinction in Spark and now I'm experiencing deprivation, whereas people who haven't used Spark might not be experiencing it. Do we have personal style sheets in N1? I'm totally willing to solve it that way if we don't want to make this change for everyone.

grinich:
Hi folks-- I suggest this be implemented as an option plugin. We're really conservative about adding color to the interface at this time, but because N1 is open source, you can change anything you want! :smile: 

More info here: https://github.com/nylas/N1/pull/799#issuecomment-166700038

colllin:
Color is only one solution.  The icons could also be redesigned, which would maintain your law of conservation of color, and help solve the design problem.

[...]

PR #799, opened 2015-12-22
Color the Trash icon a dark shade of red to make it more visually distinctive

mbilker:
Referencing #794 where @colllin suggests making the trash icon more like Spark's trash icon to prevent confusion between the Archive and Trash icons.

### DIFF
--- a/internal_packages/message-list/lib/thread-trash-button.cjsx
+++ b/internal_packages/message-list/lib/thread-trash-button.cjsx
@@ -17,7 +17,7 @@ class ThreadTrashButton extends React.Component
     focusedMailViewFilter = FocusedMailViewStore.mailView()
     return false unless focusedMailViewFilter?.canTrashThreads()
 
-    <button className="btn btn-toolbar"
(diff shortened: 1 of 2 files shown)


mbilker:
I would like feedback on how to implement this.

Could there be a switch under Preferences?
What color should it be?
What part of the button should be colored?

mbilker:
Should the icons be redesigned to be more distinctive?
Should the color change only on highlight?

grinich:
Thanks for taking the time to put this together, but it's unfortunately not something we're going to merge into master at this time. We are focused on using color very sparingly throughout the app and we only use red to indicate error during setup. Also, native Mac OS buttons don't have this type of treatment ever applied.

In the future we may re-visit using color as interface affordances. Perhaps you can make this a plugin enhancement for now instead?

mbilker:
@grinich Thank you for reviewing this! I will make a plugin for this to benefit the user.

mbilker:
@grinich I only ask that you at least give the trash icon the `btn-trash` className like how the Archive button has the `btn-archive` className.

PR #866, opened 2015-12-30
Remove log of integration test package installation by default

mbilker:
The `script/bootstrap` command appears to log the module installation by default. This PR hides the output by default using the `--no-quiet` option check.


bengotow:
Thanks @mbilker — this looks good.

PR #873, opened 2015-12-31
Make test output selectable

dcposch:
Test error messages are unselectable by default. This PR adds the ability to select and copy error messages produced by the test suite.

image


bengotow:
Looks great—thanks!

Issue #879, opened 2016-01-01
CONTRIBUTING.md incorrectly states you need an invite code


phoenixeliot:
CONTRIBUTING.md has these notes that should probably be removed, unless I'm missing something. AFAIK you no longer need an invite code.

> Note: You currently need an invitation code to use N1. You can request one here. If you reach out to our team, we can bump you to the front of the list so you can contribute to the project.

and

>The invitation flow allows us to control adoption of N1 and maintain a great experience for our users. However, the sync engine is open source and you can set it up yourself to begin using N1 immediately. Follow instructions on the sync engine repository.

and

>Once you have an invitation code:

mbilker:
Thank you for reporting this @phoenixeliot! I created the PR #880 for this issue.

PR #880, opened 2016-01-01
Remove invitation references from `CONTRIBUTING.md`

mbilker:
In response to #879, this change replaces the reference to the invitation system to the Sync Engine to notify users that contributions to N1 do not require an invitation. The build instructions only reference the email credentials are required.

I need feedback on how the first paragraph under `Contributing to N1` should be phrased.


bengotow:
Thanks—looks good!

PR #892, opened 2016-01-04
Move quick actions inside of flexbox with injectable component set

colinking:
It is difficult to extend the quick action list, since the current quick actions are inserted at once inside of the ThreadListQuickActions class. I've moved these into an injectable component set, which makes it possible to inject new quick actions that can be re-ordered to the front via a flexbox.

Here's an example:
image

### DIFF
--- a/internal_packages/thread-list/lib/thread-list-quick-actions.cjsx
+++ b/internal_packages/thread-list/lib/thread-list-quick-actions.cjsx
@@ -4,32 +4,22 @@ React = require 'react'
  TaskFactory,
  FocusedMailViewStore} = require 'nylas-exports'
 
-class ThreadListQuickActions extends React.Component
-  @displayName: 'ThreadListQuickActions'
+class ThreadArchiveQuickAction extends React.Component
+  @displayName: 'ThreadArchiveQuickAction'
   @propTypes:
     thread: React.PropTypes.object
 
   render: =>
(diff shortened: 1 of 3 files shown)


mbilker:
Please remove trailing white-space that is causing the Travis build to throw linting errors. Also, the [component register commands](https://github.com/colinking/N1/blob/d46ba0d5d5d9a3aa7f95b83091eb1c1495af5b6d/internal_packages/thread-list/lib/thread-list.cjsx#L142-L145) should not be placed inside the resolver method. It will register multiple copies of the components.

mbilker:
Looks good otherwise!

colinking:
Thanks for the feedback, I've fixed the linting issues. I looked into InjectedComponentSets and moved the default quick actions under the `children` attribute which should be a cleaner solution as compared to using the ComponentRegistry.

bengotow:
Hey @colinking — this looks great! Definitely a nice feature. Let me do a bit of profiling and make sure this won't impact the performance of the thread list too much and we can land it. By default, InjectedComponentSet wraps each child in an "UnsafeComponent", which calls `React.render` separately for it's contents. That means that a bad third-party component can't break anything outside it, but causes the React render cycle to be fragmented a bit more. Should be fine, but we can double-check.

bengotow:
Looks good! Since you moved the existing items to `children`, there's actually no performance penalty unless there are plugins adding items.

PR #899, opened 2016-01-05
Grunt `install` task: Replace `shareDir` with `linuxShareDir` to reflect global changes

mbilker:
The `install` task would fail since `linuxShareDir` was undefined for the template `.desktop` file. This is caused by the passing `--install-dir` to `script/grunt install` pointing at a path not in `/tmp` or `TMPDIR`, and this action results in a separate code path the is not normally covered under the Travis CI suite.


bengotow:
Hey @mbilker this looks great. Thanks—

PR #903, opened 2016-01-05
Copy app icons to `hicolor` on Debian

mbilker:
Debian searches the `hicolor` directory by default and the [`postinst`](https://github.com/nylas/N1/blob/master/build/resources/linux/debian/postinst#L23) script rebuilds the cache at install time. This addition will add the icons of each size to the `hicolor` directory.


bengotow:
Hey @mbilker — looks like this failed the Linux CI builds and it might be important. Can you take a look?

mbilker:
Thanks @bengotow, it was an important error. It wasn't able to copy the icon files because it was in the wrong working directory for relative paths. I dug into the `mkdeb` task and found `DEB_PATH` points to `binDir` so it should work now.

bengotow:
Sounds good—thanks!

PR #913, opened 2016-01-06
Reorder archive and trash icons 

colinking:
Because of changes made in #892, the archive and trash quick action buttons were inserted in the wrong order as can be seen below:
image
This fixes that issue so that they are in the same order as in the toolbar.

Sorry about this -- it was my fault for not checking before I submitted the previous pull request.


bengotow:
Ahh man good catch—I hadn't even noticed... thanks for the fix!

PR #926, opened 2016-01-08
[ON HOLD] Account list overhaul

mbilker:
This PR changes a whole lot. The removes the dropdown from `AccountSwitcher` and replaces the primary account with every entry from the dropdown with additions, like the active account indicator and _WIP_ unread count badge for each account.

I need feedback on **ALL** of the backend code changed. The `Observables` caused wacky errors in development sometimes with the dispose method being null during spec test runs.

**TODO**:
- Update the account switcher when the unread counts change, currently they update whenever the switcher updates and is inconsistent with the backend store

### DIFF
--- a/internal_packages/account-sidebar/lib/account-switcher.cjsx
+++ b/internal_packages/account-sidebar/lib/account-switcher.cjsx
@@ -1,7 +1,8 @@
 React = require 'react'
-{Actions, AccountStore} = require("nylas-exports")
+{Actions, AccountStore, Label, ThreadCountsStore} = require("nylas-exports")
 crypto = require 'crypto'
 {RetinaImg} = require 'nylas-component-kit'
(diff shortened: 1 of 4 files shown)


mbilker:
Looks like this will need to be redone after @jstejada finishes with his [rip-current-account](https://github.com/nylas/N1/tree/jstejada/rip-current-account).

bengotow:
Hey @mbilker — thanks for starting on this! We're currently redoing the entire account sidebar and many other things in the app to add a unified inbox. It might not be worthwhile to work on this until we're finished with that. Stay tuned!

mbilker:
@bengotow Thanks for the update. I saw that branch and interested in what changes it will bring. I still prefer multiple accounts having their own view.

mbilker:
I will reopen this if people want this after the Unified Inbox is merged in.

PR #966, opened 2016-01-12
Standardize capitalization to am/pm, for message-list

cllns:
The time format for closed vs. expanded in the message-list is not consistent.
Closed:
screen shot 2016-01-11 at 17 06 13

Expanded:
screen shot 2016-01-11 at 17 06 19

A number of other things change, but if the time format is the same, it'll appear as though the only thing changing with the time is adding the date before it.

This patch fixes that.

I have no opinion on lowercase vs. uppercase. Switching it to lowercase was just an easier change.

### DIFF
--- a/internal_packages/message-list/lib/message-timestamp.cjsx
+++ b/internal_packages/message-list/lib/message-timestamp.cjsx
@@ -28,10 +28,10 @@ class MessageTimestamp extends React.Component
       diff = now.diff(msgDate, 'days', true)
       isSameDay = now.isSame(msgDate, 'days')
       if diff < 1 and isSameDay
-        return msgDate.format "h:mm a"
+        return msgDate.format "h:mm A"
       if diff < 1.5 and not isSameDay
         timeAgo = msgDate.from now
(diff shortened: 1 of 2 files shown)


bengotow:
Hey @cllns — this looks great, thanks for the fix! How do you feel about am/pm vs AM/PM? I looked at a couple other Mac apps for comparison (Mail.app and Evernote), and they seem to capitalize them everywhere. Maybe we should standardize them in the other direction?

cllns:
I think `AM/PM` is more natural than `am/pm`, but I also kind of like the lowercase aesthetically. It's different, for sure, but it also kind of goes with the feel of N1. 

Maybe a designer at @nylas can chime in?

bengotow:
Hey @cllns — we've decided to standardize on AM / PM (capitalized). Want to update this pull request real quick and we'll land it? Thanks!

cllns:
All set @bengotow!

bengotow:
Thanks—welcome to the project! 🎉

PR #988, opened 2016-01-13
Allow custom Sync Engine specified in `config.cson`

mbilker:
In the Slack chat, `niels` and `thomas` requested this feature to use their own Sync Engines hosted on their local networks. This PR allows a custom defined Sync Engine API Root and App ID.

While the config path can change to be more isolated in its own context, I believe the one I provided is suitable.

### DIFF
--- a/src/flux/nylas-api.coffee
+++ b/src/flux/nylas-api.coffee
@@ -152,8 +152,8 @@ class NylasAPI
       @AppID = 'n/a'
       @APIRoot = 'http://localhost:5555'
     else if env in ['custom']
-      @AppID = NylasEnv.config.get('env.AppID') or 'n/a'
-      @APIRoot = NylasEnv.config.get('env.APIRoot') or 'http://localhost:5555'
+      @AppID = NylasEnv.config.get('syncEngine.AppID') or 'n/a'
+      @APIRoot = NylasEnv.config.get('syncEngine.APIRoot') or 'http://localhost:5555'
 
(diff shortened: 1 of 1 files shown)


mbilker:
I am going to test this once I get the Sync Engine running on my home server.

src/flux/nylas-api.coffee
```diff
@@ -151,6 +151,9 @@ class NylasAPI
     else if env in ['local']
       @AppID = 'n/a'
       @APIRoot = 'http://localhost:5555'
+    else if env in ['custom']
+      @AppID = NylasEnv.config.get('env.AppID') or 'n/a'
```
bengotow: Hey @mbilker — I'm not 100% sure that the config schema will allow sub-keys on `env` because it's expected to be a string. Was this working? Might need to be envAppId or another separate key.

mbilker: Oh shoot. I forgot about that. I really should test these kinds of things. I never got around to installing the Sync Engine on my home server, which doesn't have any sort of visualization support.

mbilker:
Okay @bengotow. I changed the namespace for the custom Sync Engine. I tested this with my own personal Sync Engine running on a local server in my home network (`192.168.1.x` as `syncEngine.APIRoot`). It worked and the messages the local Sync Engine pulled were shown in N1.

bengotow:
Great! Looks good. I'm going to merge this in and I'll add a note to the docs.

PR #998, opened 2016-01-13
set the ContentType when uploading packages to S3.

khamidou:
We weren't setting the correct Content-Type when uploading files to S3, which made S3 serve them with `application/octet-stream` (bad!). This diff fixes this.


mbilker:
Should the branch be deleted?

PR #1029, opened 2016-01-15
Set Phishing Detection package name to the correct value

mbilker:
This was posted by `pkortge` on the Slack chat.

The `package.json` file reported the package name as `phishing`, which is incorrect as the folder is named `phishing-detection`. This PR sets the `name` value in `package.json` to `phishing-detection` so N1 loads the package.


bengotow:
Looks good—thanks!

PR #1041, opened 2016-01-15
Allow additional parameters specified on command line

mbilker:
This prevents Electron saying it cannot find the main entrypoint, when using the `--test -s` parameters on the command line to spec test plugins.

I use this option when spec testing `email-pgp`, my PGP plugin, and want to use the N1 master repository instead of the installed N1.

### DIFF
--- a/N1.sh
+++ b/N1.sh
@@ -22,5 +22,5 @@ if [ ! -e "$ELECTRON_PATH" ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


bengotow:
Hey @mbilker —this looks good. Is this still `in-progress` or can we land it?

mbilker:
Oh. Sorry @bengotow and yes this is a go.

bengotow:
Thanks!

Issue #1011, opened 2016-01-13
On Linux, English spellchecking is disabled for `en-GB`


bengotow:
We are only detecing `en-US` for the english spelling dictionary. We should look for anything with the `en-` prefix.

mbilker:
This also depends if the Hunspell `en-GB` spelling dictionary is installed.

On my Arch Linux computer, the `en-GB` dictionaries are installed alongside the `en-US` ones:
```
[~]── - ls -lh /usr/share/hunspell/en_{US,GB}*
lrwxrwxrwx 1 root root   15 Jun  6  2015 /usr/share/hunspell/en_GB.aff -> en_GB-large.aff
lrwxrwxrwx 1 root root   15 Jun  6  2015 /usr/share/hunspell/en_GB.dic -> en_GB-large.dic
-rw-r--r-- 1 root root 3.1K Jun  6  2015 /usr/share/hunspell/en_GB-large.aff
-rw-r--r-- 1 root root 858K Jun  6  2015 /usr/share/hunspell/en_GB-large.dic
-rw-r--r-- 1 root root 3.1K Jun  6  2015 /usr/share/hunspell/en_US.aff
-rw-r--r-- 1 root root 531K Jun  6  2015 /usr/share/hunspell/en_US.dic
```

mbilker:
By default, `node-spellchecker` only contains the Hunspell `en_US` dictionary. Maybe fork `node-spellchecker` and add the other dictionaries in according to their license?

bengotow:
I'm tagging this as a good stater issue. The problem is somewhere in our implementation here: https://github.com/nylas/N1/blob/unified-inbox/src/nylas-spellchecker.coffee#L79. It'd be great to investigate what `spellchecker.getAvailableDictionaries()` returns on Linux. It might be that we're looking for `en-GB`, when the spellchecker library is returning `en_GB`, or something like that.

mbilker:
I actually have no spellchecker functionality. I do not compose many messages and do not make many mistakes, but I checked one time and I found many errors.

Here is output from the JS console:
```javascript
var Spellchecker = require('spellchecker');
Spellchecker.getAvailableDictionaries();
-> []
$n.NylasSpellchecker.getAvailableDictionaries()
-> []
```

The `$n.NylasSpellchecker.getAvailableDictionaries()` returning `[]` is odd because of that implementation.

mbilker:
[...]

PR #1145, opened 2016-01-27
Fix logic for en_US default dictionary on Linux

mbilker:
The logic uses `if not` for CoffeeScript, but this checks if the constant `"en_US"` is not null, which is not the intended function of the check.

This PR fixes the intended functionality of the function and moves the `not` call to `in`.

### DIFF
--- a/src/nylas-spellchecker.coffee
+++ b/src/nylas-spellchecker.coffee
@@ -12,7 +12,7 @@ class NylasSpellchecker
   isLanguageAvailable: (lang) =>
     return false unless lang
     dicts = @getAvailableDictionaries()
-    return (lang in dicts) or (lang.replace('_', '-') in dicts) or (lang.replace('-', '_') in dicts)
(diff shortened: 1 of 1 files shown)


mbilker:
This is for #1011

src/nylas-spellchecker.coffee
```diff
@@ -76,7 +76,7 @@ class NylasSpellchecker
   getAvailableDictionaries: ->
     if process.platform is 'linux'
       arr = spellchecker.getAvailableDictionaries()
-      if not "en_US" in arr
```
emorikawa: OMG Coffee!! Good catch!

emorikawa:
Hey @mbilker thank you for tracking this down. Can you do me a favor and rebase this on top of latest master. We just landed tons of commits from the unified inbox branch and it'll create one hell of a merge line otherwise.

mbilker:
There is the rebase. No conflicts. I will wait for Travis or AppVeyor to complete.

PR #1151, opened 2016-01-28
Check metaKey and ctrlKey presence for multi platform support

### LINKED ISSUE #388: Select multiple messages using ctrl / cmd key
Shift key works as expected in OS; click one mail in the list, shift-click another, and all mails in between are also selected.

Control/Command key should toggle selection of individual messages per click.

g105b:
Fixes #388

### DIFF
--- a/src/components/multiselect-list.cjsx
+++ b/src/components/multiselect-list.cjsx
@@ -140,7 +140,7 @@ class MultiselectList extends React.Component
 
   _onClickItem: (item, event) =>
     return unless @state.handler
-    if event.metaKey
+    if event.metaKey || event.ctrlKey
       @state.handler.onMetaClick(item)
     else if event.shiftKey
       @state.handler.onShiftClick(item)


src/components/multiselect-list.cjsx
```diff
@@ -140,7 +140,7 @@ class MultiselectList extends React.Component
 
   _onClickItem: (item, event) =>
     return unless @state.handler
-    if event.metaKey
+    if event.metaKey || event.ctrlKey
```
g105b: The `Event` object has a `ctrlKey` property, I think this should be standard on all platforms but in case I put it after the or operator as to not affect existing behaviour.

emorikawa:
@g105b This looks great I just cherry-picked on top of master and pushed. Thank you so much for contributing!

g105b:
@emorikawa Thank you. It's my first contribution to CoffeeScript, so I thought I'd pick something simple.

Could you let me know why the cherrypick method was used instead of merging the pull request? Just wondering if I did something wrong.

Issue #1176, opened 2016-02-01
Fix build modules for FreeBSD


forquare:
`%>uname -a
FreeBSD fbsd-dtop 10.2-RELEASE FreeBSD 10.2-RELEASE #0 r286666: Wed Aug 12 15:26:37 UTC 2015     root@releng1.nyi.freebsd.org:/usr/obj/usr/src/sys/GENERIC  amd64`

```
%>script/bootstrap
Node: v5.2.0
npm: v2.14.7

---> Installing N1 build tools
     This goes inside the `build` folder and runs `nam install`
     It will use the system `npm` to bootstrap our own N1 npm.
     Our build tools (like Grunt) need to be compiled against Node via `nam`.
     Everything else needs to be compiled against Chromium with `app`.

     $ npm --userconfig="/tmp/N1/.npmrc" install --loglevel error --cwd=/tmp/N1/build --ignoreStdout=true 


Traceback (most recent call last):
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/gyp_main.py", line 18, in <module>
    sys.exit(gyp.script_main())
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 533, in script_main
    return main(sys.argv[1:])
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 526, in main
    return gyp_main(args)
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 502, in gyp_main
    params, options.check, options.circular_check)
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 129, in Load
    params['parallel'], params['root_targets'])
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/input.py", line 2756, in Load
    RemoveLinkDependenciesFromNoneTargets(targets)
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/input.py", line 1501, in RemoveLinkDependenciesFromNoneTargets
    if targets[t].get('variables', {}).get('link_dependency', 0):
KeyError: 'deps/breakpad/breakpad.gyp:dump_syms#target'
gyp ERR! configure error 
gyp ERR! stack Error: `gyp` failed with exit code: 1
[...]

PR #1177, opened 2016-02-01
Bump grunt-contrib-csslint to to ~0.5.0

mbilker:
This addresses #1176. `csslint` removed the `package.json` `os` property in `0.10.0`. `grunt-contrib-csslint` updated to `csslint` in `0.2.0`. Atom uses `0.2.0`, but N1 can be better and use a newer version of the module.


emorikawa:
Thanks for tracking this down!

mbilker:
I am still working on why it fails under OS X for some reason. I am looking into a potential Babel upgrade, but that is a major change and looking to avoid that.

mbilker:
Thanks for addressing my concerns in the Slack group. I am surprised that updating `csslint` triggered the bug.

PR #1210, opened 2016-02-06
Define the remote variable

mbilker:
This code triggers when a window is supposed to close, but this error causes the DevTools to open and the window not exiting with the keyboard shortcut.

Please tell me if you want the `remote` module variable assignment within the `_.defer` block.


bengotow:
Looks good—thanks!

Issue #1301, opened 2016-02-11
Keyboard shortcut for mark-as-unread


sethvargo:
Hey there :smile: 

I started using Nylas yesterday and I'm basically in love :heart:. There's one critical feature that I'm missing (and maybe I just can't find it), but I frequently mark messages as unread when I want to come back to them later. It's a form of triage for me. I found the button in the navbar, but there's also a shortcut in apple mail and gmail (cmd + shift + U) that marks the message as unread.

Is there a way to create custom shortcuts or have this shortcut added? If you point me in the right direction, I can probably issue a PR :smile:

bengotow:
Hey! I believe you can do this if you turn on the Gmail shortcut set from Preferences > Keyboard Shortcuts. For some reason, it's actually just `Shift-U` and not `Cmd-Shift-U`. Hope that helps!

jstejada:
hey @sethvargo I'm closing this issue for now. If the gmail shortcut did not work feel free to reopen

sethvargo:
Hi @bengotow @jstejada 

Correct, but I don't use Gmail; those are not the shortcuts I'm used to. I'm migrating from Apple Mail and this does not affect the Gmail keyboard shortcuts at all. I'm not sure I understand the harm in merging this? The keyboard shortcuts match apple mail directly.

sethvargo:
(this == #1324)

PR #1324, opened 2016-02-12
More closely match Apple mail shortcuts

sethvargo:
Fixes https://github.com/nylas/N1/issues/1301


sethvargo:
Ping @bengotow

bengotow:
Looks great - thanks @sethvargo. I thought that "cmd-D" was equivalent to "cmd-shift-d" (the capital D), but we must have been wrong! We might want to check that there aren't any other places where that needs to be switched.

PR #1370, opened 2016-02-18
Toggle-button for "unseen" messages

xexpanderx:
A toggle-button that shows "unseen"-messages only would be nice.


bengotow:
Hey @xexpanderx — it doesn't seem that there's any code in this pull request?

PR #1416, opened 2016-02-25
Fix important thread labels

### LINKED ISSUE #1225: Important tag
Hi.

Since the today's update (current version v0.4.5), I have this annoying important tag that I cannot remove. I tried switching on/off in the preference. Still there, see attachment

screenshot 2016-02-07 23 16 55

patrickheeney:
To fix https://github.com/nylas/N1/issues/1225#issuecomment-188873160

### DIFF
--- a/internal_packages/thread-list/lib/thread-list-columns.cjsx
+++ b/internal_packages/thread-list/lib/thread-list-columns.cjsx
@@ -74,7 +74,10 @@ c3 = new ListTabular.Column
       ignored = [].concat(currentCategories, CategoryStore.hiddenCategories(thread.accountId))
       ignoredIds = _.pluck(ignored, 'id')
 
-      for label in (thread.sortedCategories())
+      threadLabels = thread.sortedCategories()
+      threadLabels = _.reject threadLabels, (l) -> l.name is 'important'
+
+      for label in threadLabels
         continue if label.id in ignoredIds
         c3LabelComponentCache[label.id] ?= <MailLabel label={label} key={label.id} />
         labels.push c3LabelComponentCache[label.id]
@@ -127,7 +130,10 @@ cNarrow = new ListTabular.Column
       ignored = [].concat(currentCategories, CategoryStore.hiddenCategories(thread.accountId))
       ignoredIds = _.pluck(ignored, 'id')
 
-      for label in (thread.sortedCategories())
+      threadLabels = thread.sortedCategories()
+      threadLabels = _.reject threadLabels, (l) -> l.name is 'important'
+
+      for label in threadLabels
         continue if label.id in ignoredIds
         c3LabelComponentCache[label.id] ?= <MailLabel label={label} key={label.id} />
         labels.push c3LabelComponentCache[label.id]


bengotow:
Hey! Thanks for submitting this @patrickheeney. I think this will definitely do the trick, but there might be something strange going on a bit further down.

A few lines above your changes, there's this bit that is already hiding a few other labels:

```
ignored = [].concat(currentCategories, CategoryStore.hiddenCategories(thread.accountId))
ignoredIds = _.pluck(ignored, 'id')
```

The thing is, `CategoryStore.hiddenCategories(thread.accountId)` /should/ include your account's `important` label. Can you do a bit of digging and see if you can figure out why that's not the case? That way we can be sure it'll be hidden everywhere in the app and not just in the thread list.

patrickheeney:
For me the only thing that shows up in `$n.CategoryStore.hiddenCategories(accountID)` is `All Mail, Drafts, Sent Mail`. I don't know anything about the internals at this point so I don't know what `hiddenCategories` means, but I don't have any of those categories `hidden` in anyway that I know of. It definitely doesn't include the `important` label. 

Why do the other areas of the codebase referenced in the issue, filter it out by `important` in the name?

patrickheeney:
Upon more research, here is what is going on:

[Thread list columns](https://github.com/nylas/N1/blob/master/internal_packages/thread-list/lib/thread-list-columns.cjsx#L77) returns the `sortedCategories` for the thread which adds in `important` [back in](https://github.com/nylas/N1/blob/0217e5a509ac9c475649b775d784311f33a8ab07/src/flux/models/thread.coffee#L141). I checked `$n.CategoryStore._categoryCache` and its not receiving any `important` labels from the server, so [the thread list](https://github.com/nylas/N1/blob/master/internal_packages/thread-list/lib/thread-list-columns.cjsx#L77) would not have filter out hidden categories it doesn't know about.

It would appear there is a discrepancy between the threads the categories get, and the category store.

patrickheeney:
@bengotow As we discuss on slack last week, the problem is that the API returns the `important` label from the `thread` endpoint. The `label` endpoint does not return the `important` labels. You mentioned you were going to check with the sync team on how to handle this. 

In the meantime, would you consider accepting this PR to resolve the issue now and for the next release. This PR took similar logic from other areas of the codebase (message list), so all of these areas will need to be refactored to support the new backend if it is implemented. If the sync team doesn't include the `important` label in the `labels` endpoint, then this PR would be the proposed fix.

jaymedavis:
:+1: to this PR.

patrickheeney:
It looks like N1 got re-factored today, so my PR is no longer applicable and will need to be re-visited. I put some serious hours into chasing this down this rogue bug since it popped up a month ago. I have been active in all areas I could be (multiple convos in chat, the other PRs, live debugging, etc) in hopes of assisting Nylas to bring this to a resolution ASAP. Unfortunately I have not received any response, aside from the comment in this thread. I don't have any more time to invest in contributing to Nylas at this time. Hopefully someone else can take this over. Cheers.

PR #1441, opened 2016-02-26
Rename tray icon files for Windows and Linux

### LINKED ISSUE #1422: 0.4.9 regression : System Tray
As of 0.4.9 on Windows 10 (version 1511, build 10586.104), the Tray Icon no longer appears. However, the process is still present after closing the window, creating zombie processes.

No third party plugins, 100% reproduction on my system.

This is a regression from 0.4.5 to 0.4.9, after a brief read through the commits, it may be related to 7ec30d0c7b622c70ba894a11293f896aaa8fa844, whose test plan is manual and has only been reviewed by @bengotow, who may not have Windows 10 to test the changeset on.

mbilker:
This PR fixes #1422.

In commit 7ec30d0c7b622c70ba894a11293f896aaa8fa844, the System Tray management moved to the main Electron process instead of the renderer process to account for any reloads of the main window. Unfortunately, _for some reason_, the old files names worked when loaded from the renderer process.

This fix renames the files so `nativeImage` can find them and load the images so it doesn't fail silently.


emorikawa:
@mbilker I'm verifying this now. Thank you!

emorikawa:
Thanks for the fast patch!

jstejada:
sweet!

PR #1469, opened 2016-02-27
README Update

e2corporation:
Updating Contributing guidelines to make note of the `env: custom` option so that a Sync Engine in a local environment may be accessed on an alternate IP.

Also APIRoot must contain quotes around the URL otherwise a config validation will be thrown.

Signed-off-by: Julien Chinapen <j.c@shogun.io>


bengotow:
Looks good - thanks!

Issue #1436, opened 2016-02-26
Swipe gesture for non-trackpad device


koenpunt:
For my Mac I use a Wacom tablet, and it would be nice if the swipe gestures which are now available for the trackpad, would also be supported for a tablet (or even a regular mouse) device.

I understand this has some implications, mainly in respect to dragging messages. But as I'm (almost) never dragging messages, it would like to have this behavior as an option.

bengotow:
Hi @koenpunt — is there any way to make your Wacom tablet emit "touch" events? That'd allow you to use it for gestures.

It looks like Wacom has their own touch SDK: http://www.wacomeng.com/touch/WacomFeelMulti-TouchFAQ.htm, but it's unclear whether there's any way to make their default driver act as a touch surface rather than a pointer device.

koenpunt:
My wacom doesn't support touch, but I like to make the gesture with my pen. A combination of handling javascript `mousedown` and `mousemove` should work I guess?

bengotow:
Yep—I think that's what we'd need to do. We currently support `touchstart` and `touchend` which is why I was hoping it could emit those events. If that's not possible we'll need to use the mouse events and make an option to enable swiping via mouse.

codyhatch:
I would **strongly** like an option to enable swiping via mouse.  Mailbox supported this, and I loved it for quickly managing emails with a regular desktop mouse.

I realize this might interfere with drag and drop, but I never drag and drop emails, and I love swiping.  It's a tradeoff I'd like the option to make.  Honestly, it's the only thing I miss from Mailbox.

koenpunt:
Btw, @bengotow you labeled it `platform-os-x`, but I think this also applies to windows and linux

todesignandconquer:
I echo @codyhatch's sentiment and would love being able to swipe by clicking and dragging because while swiping on a Magic Mouse works, it's very unintuitive.

PR #1480, opened 2016-02-27
add copy/paste mouse swipe support

koenpunt:
I've hacked in mouse support for message swipe actions (#1436). 

Although this does work, I think a lot of functions and properties have to be renamed to make sense for both touch and mouse.

This currently also conflicts with message dragging; maybe there should be a check for the direction of swiping, and start dragging when the "swipe" is not more-or-less horizontal.

### DIFF
--- a/src/components/swipe-container.jsx
+++ b/src/components/swipe-container.jsx
@@ -226,6 +226,39 @@ export default class SwipeContainer extends Component {
     }
   };
 
+  _onMouseDown = (e)=> {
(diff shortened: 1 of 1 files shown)


bengotow:
Hey! Thanks for submitting this, and sorry for sitting on it for... basically forever. We've decided not to extend gesture support to the mouse just yet, because we plan to do more dragging-type interactions in the future. Sorry!

PR #1506, opened 2016-02-29
Add Mail.app keybinds

sdewith:
Added jumping, search focus, select, star and archive keybinds to Apple
Mail keymap

### DIFF
--- a/keymaps/templates/Apple Mail.cson
+++ b/keymaps/templates/Apple Mail.cson
@@ -6,33 +6,26 @@
 'body':
 
   ### Jumping ###
-  'cmd-1' : 'navigation:go-to-inbox'
-  'cmd-2' : 'navigation:go-to-starred'
-  'cmd-3' : 'navigation:go-to-sent'
-  'cmd-4' : 'navigation:go-to-drafts'
-  'cmd-5' : 'navigation:go-to-all'
-  'cmd-6' : 'navigation:go-to-contacts'
-  'cmd-7' : 'navigation:go-to-tasks'
-  'cmd-8' : 'navigation:go-to-label'
+  'cmd-ctrl-1' : 'navigation:go-to-inbox'
+  'cmd-ctrl-2' : 'navigation:go-to-starred'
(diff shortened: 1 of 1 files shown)


keymaps/templates/Apple Mail.cson
```diff
@@ -4,16 +4,39 @@
...
+  'cmd-7' : 'navigation:go-to-tasks'
+  'cmd-8' : 'navigation:go-to-label'
+
+    ### Threadlist selection ###
+  'cmd-a': 'multiselect-list:select-all'
+
+
+  ### Application ###
+
+  'cmdopt-f': 'application:focus-search'
```
bengotow: I think this needs to be `cmdctrl-alt-f` to work properly. It might also be down at the bottom?

keymaps/templates/Apple Mail.cson
```diff
@@ -4,16 +4,39 @@
...
+
+
   # Triage
   'cmdctrl-r': 'application:reply'
   'cmdctrl-shift-r': 'application:reply-all'
   'cmdctrl-shift-f': 'application:forward'
   'cmdctrl-shift-j': 'application:report-as-spam'
   'cmdctrl-shift-u': 'application:mark-as-unread'
+  'cmdctrl-a': 'application:archive-item'
```
bengotow: This is a little silly, but technically `cmdctrl` maps to either "cmd" or "control" depending on the user's platform, so for the shortcut cmd+ctrl+a, it actually needs to be `cmd-ctrl-a`. (Since cmdctrl + ctrl would just be ctrl)

keymaps/templates/Apple Mail.cson
```diff
@@ -4,16 +4,39 @@
 # http://dashkards.com/img/dashkard-mail.png
 
 'body':
+
+  ### Jumping ###
+  'cmd-1' : 'navigation:go-to-inbox'
```
bengotow: Can we make these `cmd-shift-1` since cmd-1 is bound to "switch to account 1"? I don't think we can override those unfortunately.

PR #1520, opened 2016-03-01
Correct eslint errors from Travis

mbilker:
There are warnings in `spellcheck-composer-extension.es6` for the `while (true)` loops.

I do not know a better way off the top of my head, but these fixes will allow the Travis builds to complete and not fail. I am surprised my Travis builds from my personal fork still succeed.


bengotow:
Thanks looks good!

Issue #1525, opened 2016-03-01
mailto: links are tracked and don't work


ridem:
First of all, **congrats** for the link tracking feature, it's the first time that I see such a cool feature on a desktop client, neat integration!

At [https://github.com/nylas/N1/blob/master/internal_packages/link-tracking/lib/link-tracking-composer-extension.es6#L30](https://github.com/nylas/N1/blob/master/internal_packages/link-tracking/lib/link-tracking-composer-extension.es6#L30), you're converting all href attributes to tracked links. I think you should only match http* links. Getting a new browser window for mailto: links is quite bad user experience.

On top of it, the server side can't handle mailto: links, which makes the feature broken.

I can try to PR if needed.

bengotow:
Hey! Ahh good catch—that's definitely something we need to fix. If you want to take a pass at improving the regexp, pull requests are definitely welcome!

adenix:
This issue is closed, however, I am seeing this issue in Version 1.0.30 (1.0.30) on macOS Sierra Version 10.12 Beta (16A313a)

PR #1531, opened 2016-03-01
Match only valid URLs for link tracking

ridem:
As of now, link-tracking creates link tracking for every <a> tag, no matter the href attribute. This commit only creates link tracking for valid URLs.
The reason I wrote a new regex instead of rewriting linkTagRegex is that this regex is used by message-list.
https://regex101.com/r/jD5zC7/3 vs https://regex101.com/r/cK0zD8/2

Resolves: #1525


ridem:
I believe the travis-ci won't fail with commit #1520

dregitsky:
Hey @ridem, thanks for writing this! That's a really good catch with `mailto:`, we definitely don't want to replace those with redirects.

It looks like the new regex won't match `href="google.com"` though, which actually is a valid class of URLs we'd want to replace. It'd be great if the regex could match http://, https://, or no protocol.

I had a go at modifying that Gruber Regex to do this, and came up with `^\b((?:https?:\/\/|(?!\w+:\/\/))(?:[^\s()<>]+|\(([^\s()<>]+|(\([^\s()<>]+\)))*\))+(?:\(([^\s()<>]+|(\([^\s()<>]+\)))*\)|[^\s`!()\[\]{};:'".,<>?«»“”‘’]))$`
which may have subtle bugs but seems to do the trick.

ridem:
I haven't looked at it into detail, but I can't make this regex work on regex101. I get an "infinite loop" and a warning for "terrible backtracking". Any thoughts on it?

Maybe for this particular issue (tracking links) we want to keep it simple, and just exclude tags that start with "mailto" as an href:
`(<a.*?href\s*?=\s*?['"])((?!mailto).*)(['"].*?>)([\s\S]*?)(<\/a>)`

emorikawa:
@ridem, Thank you for looking into this! I just rebased your patch onto master and pushed.

I agree that ignoring `mailto` links like this is definitely a safer way to go for now. Longer term, however, I think we should use js to properly parse the urls and check for a whitelist of protocols. For the time being this will fix the immediate problem. Thank you!

PR #1576, opened 2016-03-05
CSSLint: Import "ui-variables" in theme-picker

mbilker:
This PR fixes the error from the Travis linting grunt task.

### DIFF
--- a/internal_packages/theme-picker/preview-styles/theme-option.less
+++ b/internal_packages/theme-picker/preview-styles/theme-option.less
@@ -1,3 +1,5 @@
+@import "ui-variables";
+
 html,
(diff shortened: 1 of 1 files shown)


jackiehluo:
Thanks for the PR, @mbilker! Unfortunately, that'll break the functionality of the theme picker because other imports are added conditionally after `ui-variables` to override it... I've been thinking about ways to get around it, but maybe for now we can just not have the LESS file at all and put everything into a string in the component, just to avoid the linter.

mbilker:
@jackiehluo thanks for the insight. I did not look too closely at the code and did not realize this.

PR #1588, opened 2016-03-07
Update LICENSE.md

Senasum:
Should provide an option for "(at your option)" or else delete it.


bengotow:
Hi @Senasum — the current license is the GPLv3 copied verbatim from here: http://www.gnu.org/licenses/gpl-3.0.en.html. I think that the clause "at your option" means that should a new GPL license become available, and should you want to use it, you can?

PR #1597, opened 2016-03-07
[ON HOLD] Update build dependencies

mbilker:
This PR updates the build dependencies to up to date versions indicated by `npm outdated`. I tested this with the spec suite and use this in my personal N1 fork [here](https://github.com/mbilker/N1).

I chose not to update Babel because of incompatibilities with 6.x and 5.x that need to be resolved separately. There were linter warnings and errors that are accounted for in this PR.

As usual, I would appreciate feedback on this PR and any suggestions to improve it to fix the Nylas team's requirements


mbilker:
I do not know why https://github.com/nylas/N1/pull/1597/files#diff-2005e3c4e5b1072e26b26c4072293b9dR1 is included with this PR, but the line already exists in nylas/N1@master.

bengotow:
Hey @mbilker — do you know if there are any specific advantages to using these new versions? We don't often update our dependencies unless there's a clear reason, because it can introduce bugs or other random things we don't want to deal with. We'll be moving to babel6 soonish because we want to use features recently added to ESLint, but on other stuff we've taken more of a "if it's not broken, don't fix it" sort of approach.

mbilker:
Alright then. I will leave this PR here, but I will continue testing on my own branch. Luckily nothing breaks as it stands and its only the building environment.

mbilker:
Some of the dependencies included with N1 releases do have vulnerabilities that should be addressed. I addressed them in my personal fork, and I probably will submit a PR addressing them.

mbilker:
Hey @bengotow --- Your patch to `asar` was included in v0.9. N1 uses v0.8.3 to pack N1.
```
Package                   Current  Wanted  Latest  Location
asar                       0.8.3   0.8.3  0.10.0  nylas-build
```

mbilker:
Another note is the updated `csslint` caught the `font-face` definitions in `static/email-frame.less` and complained because they should be in their own file. The 'NylasPro' font declaration does exist in `internal_packages/nylas-private-fonts/stylesheets`.

mbilker:
Well since many of the dependencies have been updated I am going to close this.

PR #1621, opened 2016-03-10
Adding ui-darkside to internal packages

jamiewilson:
Hey @jackiehluo, I've added my theme to the internal packages. I've got one question about referencing images within my theme directory. Does the following line from my CSS look correct?

https://github.com/nylas/N1/compare/master...jamiewilson:master#diff-bd40c299aac810b9205161297c26c7f9R198

Also, I'm wondering if it might make more sense to use [Git Submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) for third-party internals?


jackiehluo:
Yay! :tada: And I _think_ that we can just put the images into their own directory under `static/images`—but let me merge it and see if that works.

And we might investigate submodules in the future, but we aren't looking to it right now!

mbilker:
Submodules are not the best solution and I find them very difficult to deal with. A better solution is to use APM to fetch and install the modules using the Git repository and commit hash.

PR #1632, opened 2016-03-11
Fixes scrollbar display when content is not scrollable

### LINKED ISSUE #1008: Bug: Vertical scrollbar in sidebar
screenclip 8

As the screenshot hopefully is able to show, there now is a vertical scrollbar in the left pane where folders and the account switcher resides. Being on Windows 10 :stuck_out_tongue: .

Edit: Obviously it is not the amount of folders that provokes it. Plus I tried out both the Taiga and the default light theme, they have both the phenomenon on display ;).

DavidChouinard:
Scrollbars are currently always shown on hover, even if the content is fully fitted in the viewport, and has a non-native feel.

Fixing that requires correctly updating scrollbar state with content changes. For example, if the scrollbar is displayed (hovered) while page height is changed (e.g. a new message comes in), interacting with the scrollbar causes erratic behavior. This PR fixes that.

Fixes #1008.

screen shot

### DIFF
--- a/src/components/scroll-region.cjsx
+++ b/src/components/scroll-region.cjsx
@@ -100,6 +100,7 @@ class Scrollbar extends React.Component
     bottom: 0
     right: 0
     zIndex: 2
+    visibility: "hidden" if @state.totalHeight != 0 && @state.totalHeight == @state.viewportHeight
 
   _onHandleDown: (event) =>
     handleNode = React.findDOMNode(@refs.handle)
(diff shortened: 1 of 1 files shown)


DavidChouinard:
Not sure why the build fails so profusely — it runs fine locally?

jackiehluo:
@DavidChouinard Yeah, I tested it locally and it works fine for me! Just merged it. Thanks! :tada:

DavidChouinard:
@jackiehluo, thanks! 🙇

Issue #911, opened 2016-01-06
Menu disappears in after reopening window - Ubuntu 15.10


jstejada:
I am using Ubuntu 15.10. When I open Nylas, then close windows and open again, I will not show menu (File, Edit, View, Help). Submitted via -- https://app.intercom.io/a/apps/t7k2sjgy/inbox/unassigned/conversations/1674529239

bengotow:
This appears to be a core Electron issue. Filing a bug there with a sample app.

sholanozie:
@bengotow

Looks like this was fixed recently by the Electron team: https://github.com/atom/electron/pull/4581

bengotow:
This was unblocked on Electron and should be fixed in the most recent release. Cheers!

Issue #847, opened 2015-12-28
menu dissapears on second launch


martunta:
Ubuntu 15.04, nylas 0.32, freshly downloaded

Menu is showed only on first launch, after that there is loading cursor for a while if you move cursor over the menu, but no menu shows up. Deleting .nylas folder does not help. 

 ~~There seems to be no way to get manu back.~~

p.s. There is only this Nylas icon with menu next to the clock, that lets to do new message, open preferences and quit the program.  ~~Maybe that is somehow related?~~

update:
Looks like it is related to two different ways you can close the window. If I close complete application using nylas icon in the right and choosing 'Quit N1', and then launching it again, then menu reappears. 
If I just close Nylas N1 window by clicking x on the window corner, and then relaunch it from Dash - then no menu. 

Looks like I have to avoid using native window closing control, until it is fixed, and it will be fine

bengotow:
Hi @martunta — I think this is https://github.com/atom/electron/issues/4012. Going to close as a dup of https://github.com/nylas/N1/issues/911.

martunta:
aha. could be. thanks for looking into it. 
p.s Though as this is number 847, then 911 is the dup of this, not vice versa :)

PR #1637, opened 2016-03-11
Open links in background when ⌘ key is pressed

### LINKED ISSUE #181: Allow "background opening" of links while holding system-default modifier key (cmd/ctrl)
Most clients (that I'm familiar with, mostly on OS X) allow a link to be opened in the background (i.e. without switching application focus to the browser) while browsing an email full of links. Seems like a good idea to me!

koenpunt:
When you hold down the meta key (⌘), and click on a link in a message,
the target application will not activate, so N1 will stay in the
foreground.

NB. The update of Electron is required to support (could be `0.36.8` as well), but I noticed that the upgrade was made previously, and reverted due to issues. (#847 #911)

Electron currently only supports this for OS X: https://github.com/atom/electron/blob/master/docs/api/shell.md#shellopenexternalurl-options

Fixes #181

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/nylas/N1/issues"
   },
-  "electronVersion": "0.36.7",
+  "electronVersion": "0.36.9",
   "dependencies": {
     "async": "^0.9",
     "atom-keymap": "^6.1.1",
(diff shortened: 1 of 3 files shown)


bengotow:
Nice! Thanks for contributing this—I'm going to leave this open until we're able to resolve the other issues with the Electron upgrade, but we'll merge it as soon as that's finished.

arturogh:
Yes! So eager to get this one :+1:

bengotow:
Hey! Sorry this couldn't make it into the latest release—we'll be upgrading Electron in the next release (and doing all the necessary testing...), so we can go ahead and merge this. Thanks for contributing!

PR #1644, opened 2016-03-11
(darkside) targeting toolbars, inline svg in css, thread item refined

jamiewilson:
- Targeted the Preferences and Single panel thread list toolbars by using their parent `<span>`
- In order to make dev easier, I've embedded my images in CSS so that I don't have to reference different files between my repo and my N1 Fork
- refined the thread list item styling
- refined the thread list collapse/expand toggle

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -5,21 +5,21 @@ https://support.nylas.com/hc/en-us/sections/203638587-N1_
 Our team tries to respond to all GitHub issues. To make sure your issue is
 actionable, try to include the following information:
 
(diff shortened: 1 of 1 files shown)


bengotow:
Hey! Thanks for submitting this—we're going to try to dig through themes today and identify CSS selectors that would be useful to add to N1. Hopefully it'll allow us to make some of these things less painful and more reliable soon.

jackiehluo:
@jamiewilson I just added some classes to the toolbar that should work... Want to try them? There's now `.toolbar-RootSidebar`, `.toolbar-ThreadList`, `.toolbar-MessageList`, and `.toolbar-MessageListSidebar`, which should replace some of the `div:nth-child`'s in there!

PR #1645, opened 2016-03-11
(darkside) Fixed alignment of message headers

Eldorado234:
like the phishing indicator in the darkside theme.

Before this change: http://pasteboard.co/2bUg8PU9.png
After: http://pasteboard.co/2bUj2Bmz.png


Eldorado234:
cc @jamiewilson

bengotow:
Looks good—thanks for submitting this! I think what we should do is move the `@message-max-width` variable (800px) into our ui-variables set so that it's possible to just override that rather than needing to override the individual elements it applies to. Will see if we can do that soon.

PR #1648, opened 2016-03-11
Fixed text overflow in phishing detection indicator

Eldorado234:
Before: http://pasteboard.co/2c06q5IH.png
After: http://pasteboard.co/2c081AqZ.png


bengotow:
Thank you! This looks great.

PR #1653, opened 2016-03-11
(phishing-detection) Added button to hide phishing warning

Eldorado234:
Hi,

I just hacked a bit around to get used to React, CoffeeScript and the N1 code. I tried my best to follow the code style I recognized in the rest of the project and would really appreciate your feedback on the code I wrote.

The feature seems to work well. The only problem is the invalid plugin ID as mentioned in the code comment below.

Feel free to close the PR if you don't like the feature or the way it is implemented.

Cheers

Simon

### DIFF
--- a/internal_packages/phishing-detection/package.json
+++ b/internal_packages/phishing-detection/package.json
@@ -4,6 +4,11 @@
   "main": "./lib/main",
   "license": "GPL-3.0",
 
+  "appId": {
+    "staging": "829s742c6v8G4jqV6KHUzaWd",
+    "production": "K3Rxa487mjMP36zh7w7xvo34"
+  },
+
   "title": "Phishing Detection",
   "description": "Warns you when an email specifies a reply-to address which is not the from address.",
   "icon": "./icon.png",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,59 @@
 # N1 Changelog
 
+### 0.4.14 (3/10/16)
+
+- Features:
+
+ + Overhauled Sidebar: The sidebar now shows more accurate contact information,
+   recent conversations with the selected participant, and more.
+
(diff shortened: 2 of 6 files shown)


internal_packages/phishing-detection/package.json
```diff
@@ -4,6 +4,11 @@
   "main": "./lib/main",
   "license": "GPL-3.0",
 
+  "appId": {
+    "staging": "829s742c6v8G4jqV6KHUzaWd",
+    "production": "K3Rxa487mjMP36zh7w7xvo34"
```
Eldorado234: Generated these IDs randomly which seems to be wrong. The sync engine lets me know that it doesn't like my ID. How do I create a valid one?

bengotow:
Hey! This looks great - sorry for the delay. The only thing I'm not sure about is whether we should store the `hidden` flag on the message metadata, or make a whitelist more like the "autoload images" plugin (https://github.com/nylas/N1/blob/master/internal_packages/message-autoload-images/lib/autoload-images-store.es6). It might be nice to make it so that once you dismiss the warning, it won't be shown for the same pair of "to+replyTo" in the future... What do you think?

Also as a side note, one of our engineers is working on exposing other message headers so that we can build something more robust than just "to != replyTo", which results in a lot of false positives!

Eldorado234:
Never mind, thanks for the feedback. 

Makes totally sense to implement this as a white list. Already thought about it but wanted to start with a simple implementation to get your feedback first.

Is it the preferred way to implement data storage of plugins with a text file (like the implementation of the "autoload images" plugin) or is a database model more suitable?

Sounds awesome, good to know. Will keep this in mind when implementing the whitelist :+1:

Eldorado234:
Sorry for the long delay! Will close this PR since I unfortunately don't find the time to improve it.

PR #1764, opened 2016-03-16
Add atom-keytar for use in extensions

mbilker:
In Cypher, I store the Keybase tokens in `config.cson` currently. With the recent integration of `atom-keytar`, I can store those authentication tokens into the system keychain where they are better protected than `config.cson` just like the account tokens.


bengotow:
Hey! This looks great but could we keep the name keytar? I think it's important so that people know what the API looks like and we don't have to write documentation explaining that it's actually `keytar`.

mbilker:
Sure thing @bengotow !

PR #1781, opened 2016-03-18
Enhance issue template

steelbrain:
There were several issues with the last issue template, here are some of them

- Instructions to user that should've been comments were not, they were shown as part of the issue
- Bullets were used instead of heading, the down side is that they weren't gonna be bullets only. Users were supposed to add paragraphs below them, thus breaking the styling.
- There were no separators in place

This PR addresses these issues


grinich:
thanks!

PR #1783, opened 2016-03-18
Use `remote` exports instead of `remote.require` for dialog

mbilker:
I was inspired by atom/atom#9627 to implement similar changes to N1. I did a simple grep to find cases of `dialog` used by the renderer and main processes and then converted them to the newer style. The specs were updated as well to account for these new changes.

### DIFF
--- a/spec/stores/draft-store-spec.coffee
+++ b/spec/stores/draft-store-spec.coffee
@@ -785,16 +785,15 @@ describe "DraftStore", ->
       spyOn(NylasEnv, "isMainWindow").andReturn true
       spyOn(FocusedContentStore, "focused").andReturn(id: "t1")
       {remote} = require('electron')
-      dialog = remote.require('dialog')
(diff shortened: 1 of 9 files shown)


bengotow:
Hey @mbilker — thanks for this, is it ready to go or still in progress? Seems like a good move to me. I also wonder if switching to `remote.dialog` will fix the issue we're having with Electron 0.37 breaking all the specs. Maybe `remote.dialog` is spy-able, while `remote.require('dialog')` is not? Guess we'll see.

mbilker:
Without the second commit to fix specs, spying on it did not work. Luckily most of the main Electron modules are exported on the `require('electron')` to make it easier to use. It seems that the `remote.require` returns new objects each time.

mbilker:
Does Electron 0.37 work for you? When I tried it, the whole process would crash from a V8 abort.

PR #1790, opened 2016-03-19
Fix my mistake with AutoUpdateManager

mbilker:
I forgot that the `AutoUpdateManager` is run from the main process


mbilker:
One thing to note is that those methods are not tested by the spec suite.

PR #1807, opened 2016-03-21
[RFC] Multi-spellchecker

mbilker:
This spellchecker implementation takes cues from Slack's spellcheck implementation and scans all available Hunspell directories on startup. Unfortunately, this comes at a startup penalty to all N1 windows (main, composer, and background). I have not tested running the spellchecker from the main process and communicating via Electron `ipc`.

This is **NOT** tested on OS X and Windows and I need to write specs for it, but I want feedback on future implementation and how to test the Windows version.


mbilker:
I could also add a preferences option to manually select the language or use [cld](https://www.npmjs.com/package/cld), which is what Slack uses to detect the language and switch the spellchecker on the fly.

mbilker:
This adds about 1 second to startup time, I am still trying to bring it down. Luckily with the new lazy load from @emorikawa it does not startup in other windows when not needed.

bengotow:
Hey @mbilker—hope this finds you well! If you're interested in carrying this forward, we're super interested in using this library, which was recently developed over at Slack: https://github.com/paulcbetts/electron-spellchecker

mbilker:
Alright! I will be sure to check it out and implement it.

bengotow:
Hey thanks for your work on this - we've landed multi-spellchecker support (powered by a library from Slack) in v0.4.59, so I'm gonna go ahead and close this.

PR #1844, opened 2016-03-30
Color Emoji on Linux through Noto Color Emoji

mbilker:
**Description:**
The Noto Color Emoji is a popular TrueType font that provides the color emojis in use on Android and is available with a very open license. N1 lacks good emoji support on Linux because it depends on an emoji font. This PR integrates the Noto Color Emoji font (2.5 MiB) into N1 so Linux users can enjoy colorful emojis.

**RFC:**
This can be applied to Windows, but I am not aware of emoji support on Windows. The last time I checked a few months ago, it was spotty and some applications used a fallback to Arial, which looks awful.

If Windows does not have great emoji support then this change can be enabled for Windows.

I do experience errors in N1 if I do not have the Noto Color Emoji TTF installed as a system font or into the user's `~/.fonts` folder. I tested this with N1 running under Electron v0.36.12.


mbilker:
Also, tutorials on creating `@font-face` declarations vary. The `font-family` in CSS may need to be changed to not collide with the system installed Noto Color Emoji font.

From my experience, it did not cause issue that I installed the Noto Color Emoji font globally on my computer.

bengotow:
Hey @jackiehluo — is this necessary given the work you're doing? @mbilker I'm not sure we'll be sticking with an emoji font. We may switch to doing them the way Slack, etc. do and use a PNG spritesheet or image set within the app, and swap them out when they're sent. We've had a lot of problems with emoji randomly missing on Windows, a handful being available on Mac OS X 10.11 but not 10.10, etc.

mbilker:
That sounds like a better plan. I know that Discord uses [Twitter emoji](https://github.com/twitter/twemoji) in some form. Slack gives the option to select out different sets and the code currently supports a replacement using images, which is more often going to have more consistency across platforms.

Again, I created this PR as a possible fix and am okay with a rejection.

jackiehluo:
Ahhh, yeah! I'll try to add them ASAP. We just have to decide on a good option; @sdewith has some ideas, I think.

PR #1898, opened 2016-04-06
fix(composer-emoji): Switch to characters to PNGs

jackiehluo:


internal_packages/composer-emoji/lib/emoji-composer-extension.jsx
```diff
@@ -205,17 +206,13 @@ class EmojiComposerExtension extends ComposerExtension {
...
-                      src="images/composer-emoji/missing-emoji/${emojiName}.png"
-                      width="14"
-                      height="14"
-                      style="margin-top: -5px;">`;
-      editor.insertHTML(html, {selectInsertion: false});
-    } else {
-      editor.insertText(emojiChar);
-    }
+    const html = `<img
+                    class="emoji ${emojiName}"
```
bengotow: It might be nice to unify this HTML block with the one above—this one references `getImagePath`, while the other one has it hardcoded for emojiName, etc. Maybe a helper method?

internal_packages/composer-emoji/lib/emoji-composer-extension.jsx
```diff
@@ -133,16 +134,16 @@ class EmojiComposerExtension extends ComposerExtension {
 
   static applyTransformsToDraft = ({draft}) => {
     const nextDraft = draft.clone();
-    nextDraft.body = nextDraft.body.replace(/<img class="missing-emoji ([a-zA-Z0-9-_]*)" [^<]+>/g, (match, emojiName) =>
-      `<span class="broken-emoji ${emojiName}">${emoji.get(emojiName)}</span>`
+    nextDraft.body = nextDraft.body.replace(/<img class="emoji ([a-zA-Z0-9-_]*)" [^<]+>/g, (match, emojiName) =>
+      `<span class="emoji ${emojiName}">${emoji.get(emojiName)}</span>`
```
bengotow: Is there any way to match the emoji characters themselves with a RegExp? It'd be kind of cool if these functions went from emoji `` => emoji unicode character and back, so that we could handle drafts created in other apps that have unicode character emoji, pasting, etc.

internal_packages/composer-emoji/lib/emoji-composer-extension.jsx
jackiehluo: Oh, yeah, I already changed that! It should be `getImagePath`. Easier to determine the platform and send the correct file path from `EmojiStore`, I think, instead of doing it across three or four different files.

PR #1907, opened 2016-04-08
fix(imports): switch to Electron's require('electron')

mbilker:
Electron in 0.37.5 phases out the usage of `require('built-in-module')` in favor of `require('electron').builtInModule`. This commit corrects usage in some cases that cause N1 to not start under Electron 0.37.5.

### DIFF
--- a/internal_packages/composer-templates/spec/template-store-spec.es6
+++ b/internal_packages/composer-templates/spec/template-store-spec.es6
@@ -1,7 +1,9 @@
 import fs from 'fs';
-import shell from 'shell';
+import { remote } from 'electron';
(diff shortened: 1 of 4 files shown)


bengotow:
:+1: let me know when this is ready to go — definitely looking to move to 0.37.5 now that the IPC crash from 0.37.4 is fixed and the N1 release is (almost) out the door.

mbilker:
Forgot to remove the label. Yes 0.37.5 is much more stable than the earlier 0.37 versions. I am using it for N1 now and aside from N1 bugs it works perfect. I still need to trace back these `atom.asar` bugs. I probably need to do a large `grep` to find `require` calls to the internal modules.

PR #1922, opened 2016-04-10
24 h clock old

ForNeVeR:


ForNeVeR:
Sorry, I've pushed the wrong button :(

PR #1522, opened 2016-03-01
Added 24hr Time

ChrisKatsaras:
My team and I were successfully able to implement the 24hr clock feature. Let me know if any changes are needed for this feature. 
-Chris

Issue #1917, opened 2016-04-09
Custom time and date format


Cu8eR:
Custom time and date format setting wuld be really useful 

DD/MM/YYYY; MM/DD/YYYY; 12/24 time format; DD/MM etc

rudolphpienaar:
+100

PR #1923, opened 2016-04-10
Add 24-hour clock support

### LINKED ISSUE #682: Feature Request: Support 24-hour / military time
It would be  great to be able to configure the time format from 12:00 to 24:00 hr. So that 11pm is 23:00Hr in Europe. Military time as my wife calls it.... (We are not that military here in Europe tho) :)) And additionally for calendar purposes, to be able to switch the first day of the week from monday to sunday.

ForNeVeR:
I've based my work on #1522, but rewritten everything because there was a strange branch history (guys there have reimported the whole N1 repository from scratch, and merged it with the main branch afterwards) and it was not fully functional.

So I'm the only copyright holder for this contribution, and I've signed CLA, so all the legal issues're covered. Please review this and let me know if any additional changes are required.

It will close #682.

### DIFF
--- a/src/date-utils.es6
+++ b/src/date-utils.es6
@@ -150,6 +150,22 @@ const DateUtils = {
     }
     return moment(date)
   },
+
+  getTimeFormat(opts) {
+    const use24HourClock = NylasEnv.config.get('core.workspace.use24HourClock')
+    let timeFormat = use24HourClock ? "HH:mm" : "h:mm"
+    if (opts && opts.seconds) {
+      timeFormat += ":ss"
+    }
+    if (!use24HourClock && opts && opts.upperCase) {
+      timeFormat += " A"
+    }
+    if (opts && opts.timeZone) {
+      timeFormat += " z"
+    }
+
+    return timeFormat
+  },
 };
 
 export default DateUtils
--- a/internal_packages/events/lib/event-header.cjsx
+++ b/internal_packages/events/lib/event-header.cjsx
@@ -3,6 +3,7 @@ path = require 'path'
 React = require 'react'
 {RetinaImg} = require 'nylas-component-kit'
 {Actions,
+ DateUtils,
  Message,
  Event,
  Utils,
@@ -44,6 +45,7 @@ class EventHeader extends React.Component
     @_unlisten?()
 
   render: =>
+    timeFormat = DateUtils.getTimeFormat timeZone: true
     if @state.event?
       <div className="event-wrapper">
         <div className="event-header">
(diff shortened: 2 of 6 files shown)


ForNeVeR:
It seems that the Travis failure has been happened due to some npm problems. Could you please restart it?

Cu8eR:
what about this?
https://github.com/nylas/N1/issues/1917

ForNeVeR:
@Cu8eR, there are *many* date and time formats in the N1 codebase. We either need to define a separate setting for every one of them, or somehow unify them to maybe use single date and time format everywhere. We need to think about it a little bit more, and maybe we need someone more experienced with N1 codebase than me to implement that.

bengotow:
Hey folks! Sorry for the delay on this. I agree there are currently a lot of time/date format strings, and we haven't put enough effort into unifying them. It'd also be really great to avoid adding a setting for this, and instead match the system settings. I was hoping that the new `Intl.DateTimeFormat().resolvedOptions()` in Chromium would allow us to retrieve the user's preference, but it only returns the time zone, numbering system, locale, etc. The `hour12` property is unset...

ForNeVeR:
I agree with you, it would be much better to get the setting from the system-wide configuration instead of messing with N1 settings. Maybe more future-proof solution would be to open an issue at the Chromium bug tracker? Are the guys there know that their implementation lacks of `hour12` option?

src/date-utils.es6
```diff
@@ -150,6 +150,22 @@ const DateUtils = {
     }
     return moment(date)
   },
+
+  getTimeFormat(opts) {
```
rpearce: If you've got parameter destructuring available to you, I reckon you could clean this code up a bit like so:
```js
getTimeFormat({ seconds, timeZone, upperCase }) {
  const use24HourClock = NylasEnv.config.get('core.workspace.use24HourClock')
  let timeFormat = use24HourClock ? "HH:mm" : "h:mm"
  if (seconds) {
    timeFormat += ':ss'
  }
  if (!use24HourClock && upperCase) {
    timeFormat += ' A'
  }
  if (timeZone) {
    timeFormat += ' z'
  }

  return timeFormat
}
```
While there's still a number of conditionals that could be pulled out and made in to a set of functions that are applied to the original string, but hey why overcomplicate.

rpearce: Sorry for dropping this! Just read the rest of the comments >.<

ForNeVeR: Anyways that's a good suggestion, thank you. I think I'll implement it.

Bitbored:
Any chance this will get merged?

bengotow:
Hey folks! Sorry for the delay getting this merged. This looks really nice—I think inserting just the "time" component of the different format strings based on the setting helped contain the scope of this quite a bit.

Going to flatten and merge this morning! We're working on a major release of N1, so it may be another week before it makes it into a public version. Cheers—

bengotow:
Landed in 899193908ed5ea36dbae8e59757c0181484629f1. Cheers!

ForNeVeR:
Thank you, Ben!

PR #1925, opened 2016-04-10
Expose 'react-addons-test-utils' for spec testing external plugins

mbilker:
In Cypher, I use `ReactTestUtils` and was exposed as `React.addons.ReactTestUtils` in React 0.13, but no longer in React 0.14. To run spec tests properly, Cypher needs `ReactTestUtils` exposed.

This PR exposes it through `nylas-exports`, but it has the side-effect of exposing it while not in dev mode. If there is a better way to expose this during spec mode, please feel free to comment.


bengotow:
Looks good to me—thanks!

PR #1933, opened 2016-04-11
Added codeclimate integration

ghoshnirmalya:
Thought that it would be a nice idea to let Codeclimate (https://codeclimate.com)handle the lints.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,24 @@
+engines:
+  brakeman:
+    enabled: true
+  bundler-audit:
+    enabled: true
+  csslint:
+    enabled: true
+  coffeelint:
+    enabled: true
+  duplication:
+    enabled: false
+  eslint:
+    enabled: true
+  fixme:
+    enabled: true
+  phpmd:
+    enabled: false
+  radon:
+    enabled: false
+  rubocop:
+    enabled: false
+exclude_paths:
+- docs/**/*
+- examples/**/*


bengotow:
Hey! I don't have much experience with CodeClimate—is there anything in particular you like it for? I took a look at the metrics it generated for you (https://codeclimate.com/github/ghoshnirmalya/N1/issues) but it seems like there's a lot of noise. We currently run `eslint` as part of the Travis flow (and it actually breaks the builds!) so I think we ensure that PRs and master are always "A" quality code.

ghoshnirmalya:
The thing I like about codeclimate is that it uses csslint, coffeelint, fixme along with eslint. So, this will take care of other things including eslint and in a way, help to make the code better. I thought that this would help in the long run with the hope that when more people are collaborating, this will take care of all the standards.

About the issues which cropped up in the result, we can tweak the codeclimate integration file (right now, we are using the default configuration file) so that we can ignore specific portions of the application where we don't need codeclimate to run :)

If you guys think that this is a nice idea and would help in standardising the code quality, we can use it :)

bengotow:
Hey! Sorry for the delay - we've talked about this internally and decided to hold off on Codeclimate until we can configure it to make it more valuable. I think the issues view (https://codeclimate.com/github/ghoshnirmalya/N1/issues) leaves a lot to be desired, but we may just need to create a suitable config file for it. We're not sure the other metrics (LOC, "churn vs code quality") are meaningful for us, and we run a wide array of linters during the Travis builds.

Thanks for bringing this up, though—hadn't heard of Codeclimate!

PR #1943, opened 2016-04-13
`nylas-env-spec` handle remote method calls through `electron.remote`

mbilker:
This PR fixes the rather cryptic error during https://travis-ci.org/nylas/N1/jobs/122675741. I experienced a lot of these trying to work patch up N1 to work under Electron 0.37.


bengotow:
Thanks! Beat me to it :-)

PR #1994, opened 2016-04-18
Expose ParticipantsTextField through component kit

DakotaNelson:
Thus enabling it to be used by anyone. This might need a styling pass in the future - it requires some slightly sketchy CSS right now to keep it from being very specifically geared toward placing recipients in an email draft.


bengotow:
Looks good! Let's land it.

PR #2004, opened 2016-04-19
Doing write on a GET is invalid.

KevinMartin:
Ending the request sends it to the server (without an invalid GET data body).

Tested with 0.4.19 and 0.4.25.


bengotow:
Hy @KevinMartin — does calling `end` close the connection? We want to leave the connection open for as long as possible to receive data from the streaming API.

KevinMartin:
No, not at all. `.end()` finalizes the request, initiating the response.

Node documentation: https://nodejs.org/api/http.html#http_http_request_options_callback

bengotow:
Cool! Looks good then—

PR #2011, opened 2016-04-20
feat(keybase): Initial commit from n1-keybase repo

DakotaNelson:


internal_packages/keybase/lib/decrypt-button.cjsx
```diff
@@ -0,0 +1,66 @@
...
+        <div className="message" ref="message">{@state.status}</div>
+      </div>
+    else if @state.wasEncrypted and @state.isDecrypted
+      # TODO a message saying "this was decrypted with the key for ___@___.com"
+      <div className="n1-keybase">
+        <div className="decrypted" ref="decrypted">{@state.status}</div>
+      </div>
+    else
+      # TODO inform user of errors/etc. instead of failing without showing it
+      <div className="n1-keybase">
```
bengotow: Would it be sufficient to show state.status here?

DakotaNelson: IIRC the issue here is that status messages from previous decryptions/failures/etc. will persist. We considered a timeout, but it then becomes sort of unclear what the timeout value should be.

DakotaNelson: (Although, you're right, message status is very poorly handled in general.)

internal_packages/keybase/lib/decrypt-button.cjsx
```diff
@@ -0,0 +1,66 @@
...
+
+  _onClick: =>
+    {message} = @props
+    passphrase = ReactDOM.findDOMNode(@refs.passphrase).value
+    for recipient in message.to
+      # right now, just try to unlock all possible keys
+      # (many will fail - TODO?)
+      privateKeys = PGPKeyStore.privKeys(address: recipient.email, timed: false)
+      for privateKey in privateKeys
+        PGPKeyStore.getKeyContents(key: privateKey, passphrase: passphrase)
```
bengotow: I think it might be good to wrap this in a convenience method and move it to the PGPKeyStore, and then have this React component call an action, like `KeybaseActions.decryptMessage(message, passphrase)`

internal_packages/keybase/lib/encrypt-button.cjsx
```diff
@@ -0,0 +1,140 @@
...
+    @state.plaintext = ""
+
+    # store the message's body here, for comparison purposes (so that if the
+    # user edits an encrypted message, we can revert it)
+    @state.cryptotext = ""
+
+  componentDidMount: ->
+    @unlistenKeystore = PGPKeyStore.listen(@_onKeystoreChange, @)
+    # periodically check to see if we should get any keys
+    @interval = setInterval(@_onTrigger, 100)
```
bengotow: We actually changed this earlier today so that your `EncryptMessageButton` will receive a "draft" prop, and you can implement `componentWillReceiveProps` to see if there are any new participants. Ping me and we can try implementing it.

internal_packages/keybase/lib/encrypt-button.cjsx
```diff
@@ -0,0 +1,140 @@
...
+
+  componentWillUnmount: ->
+    @unlistenKeystore()
+    clearInterval(@interval)
+
+  _getStateFromStores: ->
+    keys = []
+    DraftStore.sessionForClientId(@props.draftClientId).then (session) =>
```
bengotow: You can actually get this session off of `props.session` now!

internal_packages/keybase/lib/encrypt-button.cjsx
```diff
@@ -0,0 +1,140 @@
...
+  _onKeystoreChange: =>
+    if @state.currentlyEncrypted
+      # message is no longer encrypted with all keys - decrypt it!
+      @_toggleCrypt()
+    # now actually get the new keys
+    @setState(@_getStateFromStores())
+
+  _onTrigger: =>
+    # Some sort of outside event has occured that we must respond to
+    DraftStore.sessionForClientId(@props.draftClientId).then (session) =>
```
bengotow: I think we can safely move this to `componentWillReceiveProps`

internal_packages/keybase/lib/encrypt-button.cjsx
```diff
@@ -0,0 +1,140 @@
...
+          session.changes.add({body: body})
+      else
+        plaintext = session.draft().body
+        @setState({plaintext: plaintext})
+
+    @setState(@_getStateFromStores())
+
+  render: ->
+    <div className="n1-keybase">
+      <button title="Encrypt email body" className="btn btn-toolbar" onClick={ => @_onClick()} ref="button">
```
bengotow: We should be able to make this `onClick={@_onClick}`. Making an extra little bound function that calls onClick shouldn't be necessary.

internal_packages/keybase/lib/encrypt-button.cjsx
```diff
@@ -0,0 +1,140 @@
...
+
+    # remove the nulls
+    kms = _.reject(kms, (key) -> key == null)
+    params =
+      encrypt_for: kms
+      msg: text
+    pgp.box params, (err, result_string, result_buffer) =>
+      if err
+        console.warn err
+      DraftStore.sessionForClientId(@props.draftClientId).then (session) =>
```
bengotow: Doesn't look like we need the session here!

internal_packages/keybase/lib/encrypt-button.cjsx
```diff
@@ -0,0 +1,140 @@
...
+        @setState({currentlyEncrypted: false})
+      else
+        # if not encrypted, save the plaintext, then encrypt
+        plaintext = session.draft().body
+        @_encrypt(plaintext, @state.keys)
+        @setState({currentlyEncrypted: true, plaintext: plaintext})
+
+  _encrypt: (text, keys) =>
+    # addresses which don't have a key
+    nullAddrs = _.pluck(_.filter(@state.keys, (key) -> return key.key is null), "address")
```
bengotow: It seems like we're constantly maintaining a set of keys on `state`. What if we just fetched the keys when you click encrypt instead?

bengotow: `key.key` is a bit strange - is `state.keys` really keys, or would another name make more sense?

internal_packages/keybase/lib/encrypt-button.cjsx
```diff
@@ -0,0 +1,140 @@
...
+    if nullAddrs.length > 0
+      missingAddrs = nullAddrs.join('\n- ')
+      # TODO this message is annoying, needs some work
+      # - link to preferences page
+      # - formatting, probably an error dialog is the wrong way to do this
+      # - potentially an option to disable this warning in the pref. page?
+      NylasEnv.showErrorDialog("At least one key is missing - the following recipients won't be able to decrypt the message:\n- #{missingAddrs}\n\nYou can add keys for them from the preferences page.")
+
+    # remove the nulls
+    kms = _.reject(kms, (key) -> key == null)
```
bengotow: Technically you can use `_.compact`!

internal_packages/keybase/lib/encrypt-button.cjsx
```diff
@@ -0,0 +1,140 @@
...
+    params =
+      encrypt_for: kms
+      msg: text
+    pgp.box params, (err, result_string, result_buffer) =>
```
bengotow: Sort of a tiny nit, but it'd be nice to use camel case here

internal_packages/keybase/lib/key-manager.cjsx
```diff
@@ -0,0 +1,35 @@
...
+      if key.key?
+        uid = "key-manager-" + key.key.get_pgp_fingerprint().toString('hex')
+      else if key.keybase_user?
+        uid = "key-manager-" + key.keybase_user.components.username.val
+      else
+        uid = "key-manager-" + key.addresses.join('')
+      return <KeybaseUser profile={key} key={uid} />
+    )
+
+    if not keys? or keys.length < 1
```
bengotow: I don't think there should ever be a scenario where `not keys?` since _.map will always return an array!

internal_packages/keybase/lib/key-manager.cjsx
```diff
@@ -0,0 +1,35 @@
...
+  @propTypes:
+    keys: React.PropTypes.array.isRequired
+
+  constructor: (props) ->
+    super(props)
+
+  render: ->
+    {keys} = @props
+
+    keys = _.map(keys, (key) =>
```
bengotow: This isn't super important, but we're trying to move toward using the built-in methods in ES6 `array.map`, `array.forEach` and `array.filter` —the underscore versions are a littler slower.

internal_packages/keybase/lib/keybase-search.cjsx
```diff
@@ -0,0 +1,68 @@
...
+      loading: false
+    }
+
+    @debouncedSearch =  _.debounce(@_search, 200)
+
+  _search: ->
+    if @state.query != "" and @state.loading == false
+      @setState({loading: true})
+      kb.autocomplete(@state.query, (profiles) =>
+        if profiles?
```
bengotow: Just from a minor styling standpoint, it'd be sort of nice to unify these code paths like this:

```
@setState({
   results: (profiles || []).map (profile) -> {keybase_user: profile},
   loading: false,
})


```

internal_packages/keybase/lib/keybase-search.cjsx
```diff
@@ -0,0 +1,68 @@
...
+      profiles = []
+
+    if @state.loading
+      loading = "LOADING"
+    else
+      loading = null
+
+    <div className="keybase-search">
+      <div className="searchbar">
+        <input type="text" placeholder="search keybase" ref="searchbar" onChange={@_queryChange} />
```
bengotow: It'd be good to get in the habit of prefixing these callbacks with `_on`, eg `_onQueryChange`

internal_packages/keybase/lib/keybase-user.cjsx
```diff
@@ -0,0 +1,237 @@
+{Utils, React, Actions, ReactDOM} = require 'nylas-exports'
+{ParticipantsTextField} = require 'nylas-component-kit'
+PGPKeyStore = require './pgp-key-store'
+kb = require './keybase'
+_ = require 'underscore'
+
+class EmailPopover extends React.Component
```
bengotow: Let's move this thing to a separate file just for the sake of it :)

internal_packages/keybase/lib/keybase-user.cjsx
```diff
@@ -0,0 +1,237 @@
...
+      </button>
+      )
+
+    # email addresses
+    if profile.addresses?.length > 0
+      emails = _.map(profile.addresses, (email) =>
+        # TODO make that remove button not terrible
+        return <li>{ email } <small><a onClick={ @_removeEmailClick } data-email={email}>(X)</a></small></li>)
+
+      if @state.inputEmail
```
bengotow: I'm not 100% sure I follow the inputEmail boolean on state. Maybe we could express this in a simpler way? "If emails.length > 0, show list, else show button?"

internal_packages/keybase/lib/keybase-user.cjsx
```diff
@@ -0,0 +1,237 @@
...
+    else
+      saveDeleteButton = (<button title="Delete" className="btn btn-toolbar btn-danger" data-email={ profile.addresses[0] } onClick={ @_delete } ref="button">
+        Delete Key
+      </button>
+      )
+
+    # email addresses
+    if profile.addresses?.length > 0
+      emails = _.map(profile.addresses, (email) =>
+        # TODO make that remove button not terrible
```
bengotow: 💄💯

internal_packages/keybase/lib/keybase-user.cjsx
```diff
@@ -0,0 +1,237 @@
...
+    else
+      # can't possibly have them saved if we have no email for them
+      saved = false
+    if not saved
+      saveDeleteButton = (<button title="Import" className="btn btn-toolbar" onClick={ @_importKey } ref="button">
+        Import Key
+      </button>
+      )
+    else
+      saveDeleteButton = (<button title="Delete" className="btn btn-toolbar btn-danger" data-email={ profile.addresses[0] } onClick={ @_delete } ref="button">
```
bengotow: This would actually be a nice place to use a bound function. You could say:

```onClick={ => @_onDeleteEmail(profile.addresses[0]) } ```

And not need to attach the email to the react element's dataset. We do that occasionally, but it's a bit hairy.

File: internal_packages/keybase/lib/keybase-user.cjsx [hunk 19]
```diff
@@ -0,0 +1,237 @@
...
+
+  _delete: (event) =>
+    # delete a locally saved key
+    email = event.target.attributes['data-email'].value
+    keys = PGPKeyStore.pubKeys(email)
+    if keys? and keys.length == 1
+      key = keys[0]
+    else if keys.length > 1
+      # TODO key metadata objects need a unique ID
+      key = @_matchKeys(@props.profile.key, keys)
```
bengotow: Would it be safe to call `matchKeys` all the time, even if there's just one? Sort of silly, but I feel like having two code paths could mean we accidentally don't test `matchKeys` :-)

internal_packages/keybase/lib/keybase-user.cjsx
```diff
@@ -0,0 +1,237 @@
...
+      #console.error "Multiple keys associated with #{email}, unable to determine which to remove from"
+
+    if key?
+      PGPKeyStore.deleteKey(key)
+    else
+      console.error "Unable to fetch key for #{email}"
```
bengotow: It'd probably be OK to surface this to the user in an error message box (`NylasEnv.showErrorDialog("message")`)

internal_packages/keybase/lib/keybase-user.cjsx
```diff
@@ -0,0 +1,237 @@
...
+  _addEmailInput: (contacts) =>
+    # when a new email is added to the list of emails
+    # flow is (click "add email") -> _addEmailClick -> (enter email) -> this
+    emails = _.pluck(contacts.to, 'email')
+    _.each(emails, @_addEmail)
+    @setState({inputEmail: false})
+
+  _addEmailClick: (event) =>
+    # create a text field in which to enter a new email to associate with a key
+    @setState({inputEmail: true})
```
bengotow: Ahh I see what we're doing here! Ignore my previous comment about inputEmail :-)

internal_packages/keybase/lib/keybase-user.cjsx
```diff
@@ -0,0 +1,237 @@
...
+    # TODO focus on the new field
+    # React.findDOMNode(@refs.addNewEmail).focus()
+
+  _removeEmailClick: (event) =>
+    email = event.target.attributes['data-email'].value
+    @_removeEmail(email)
+
+  _removeEmail: (email) =>
+    PGPKeyStore.removeAddressFromKey(@props.profile, email)
```
bengotow: I think it'd be OK to merge this with the function above. It'd also be good to switch to using Actions instead of directly calling methods on the store. Usually packages create actions like this: https://github.com/nylas/N1/blob/wip/keybase/internal_packages/thread-snooze/lib/snooze-actions.js#L1, and then the store listens to the action and performs the change.

internal_packages/keybase/lib/keybase-user.cjsx
```diff
@@ -0,0 +1,237 @@
...
+          </small>
+        </h2>
+
+        { profileList }
+      </div>)
+
+    # button to save/delete key
+    if profile.addresses?.length > 0
+      # TODO what about the other email(s) that might be in the list?
+      saved = PGPKeyStore.pubKeys(profile.addresses[0])
```
bengotow: We should technically call this method when the react component gets it's `@props.profile` and save it to state.

internal_packages/keybase/lib/keybase.coffee
```diff
@@ -0,0 +1,64 @@
...
+request = require 'request'
+
+class KeybaseAPI
+  constructor: ->
+    @baseUrl = "https://keybase.io"
+
+  getUser: (key, keyType, cb) =>
+    if not keyType in ['usernames', 'domain', 'twitter', 'github', 'reddit',
+                       'hackernews', 'coinbase', 'key_fingerprint']
+      console.error 'keyType must be a supported Keybase query type.'
```
bengotow: I really like this being here—it's always a good idea to add assertions to things like this. Maybe it should just `throw new Error("bla")`? If it's really a "this is not going to work" sort of thing.

internal_packages/keybase/lib/pgp-key-store.cjsx
```diff
@@ -0,0 +1,391 @@
...
+
+    @_pubWatcher = null
+    @_privWatcher = null
+
+    @_keyDir = path.join(NylasEnv.getConfigDirPath(), 'keys')
+    @_pubKeyDir = path.join(@_keyDir, 'public')
+    @_privKeyDir = path.join(@_keyDir, 'private')
+
+    # Create the key storage file system if it doesn't already exist
+    fs.access(@_keyDir, fs.R_OK | fs.W_OK, (err) =>
```
bengotow: This looks great—I appreciate that it's all async (so it won't slow down the window loading.)

It might be interesting to consider the failure cases. For example, access + mkdir is nice and safe, but we could also try to mkdir all the time, and ignore the error if it already exists (since mkdir is non-destructive!)

internal_packages/keybase/lib/pgp-key-store.cjsx
```diff
@@ -0,0 +1,391 @@
...
+          if err
+            fs.mkdir(@_pubKeyDir, (err) =>
+              if err
+                console.warn err))
+        fs.access(@_privKeyDir, fs.R_OK | fs.W_OK, (err) =>
+          if err
+            fs.mkdir(@_privKeyDir, (err) =>
+              if err
+                console.warn err))
+        @_populate(isPub = true)
```
bengotow: I'm not actually sure the assignment in here is doing what you expect (this is python :-)). I think this is essentially being rewritten as:
```
isPub = true
@_populate(isPub)
```

internal_packages/keybase/lib/pgp-key-store.cjsx
```diff
@@ -0,0 +1,391 @@
...
+        @_populate(isPub = false)
+        @watch())
+
+    @unlisten = DraftStore.listen(@_onDraftChanged, @)
+
+  validAddress: (address, isPub) =>
```
bengotow: this is really minor, but I think the JS convention is to say `isValidAddress`

internal_packages/keybase/lib/pgp-key-store.cjsx
jackiehluo: > this is python :-)
been there 😪

internal_packages/keybase/lib/pgp-key-store.cjsx
```diff
@@ -0,0 +1,391 @@
...
+  _populate: (isPub) =>
+    # add metadata elements (sans keys) to later be populated with actual keys
+    # from disk
+    if isPub
+      keyDirectory = @_pubKeyDir
+      @_pubKeys = []
+    else
+      keyDirectory = @_privKeyDir
+      @_privKeys = []
+    fs.readdir(keyDirectory, (err, filenames) =>
```
bengotow: It' be really nice to consider making a class to represent these "key file" hashes ({path, addresses}) so that we could change the implementation totally within that class in the future. I imagine this could look something like this:

```
fs.readdir(keyDirectory, (err, filenames) =>
   filenames.filter((name) => name isnt '.').map((name) => {
       new SavedKey(path.join(keyDirectory, filename))
   })


```

internal_packages/keybase/lib/pgp-key-store.cjsx
```diff
@@ -0,0 +1,391 @@
...
+          @_displayError(err)
+      )
+
+  deleteKey: (key) =>
+    if this._displayDialog(
+        'Delete this key?',
+        'The key will be permanently deleted.',
+        ['Delete', 'Cancel']
+    )
+      fs.unlink(key.path, (err) =>
```
bengotow: It'd be nice to move this fs read/write stuff into a `SavedKey` class as well

internal_packages/keybase/lib/pgp-key-store.cjsx
```diff
@@ -0,0 +1,391 @@
...
+          key.timeout > Date.now()
+      else
+        keys = address_keys
+    return keys
+
+  _displayError: (message) ->
+    dialog = remote.require('dialog')
+    dialog.showErrorBox('Key Management Error', message)
+
+  _displayDialog: (title, message, buttons) ->
```
bengotow: It might be good to remove these in favor of `NylasEnv.showOpenDialog`, which we may have failed to mention :-)

internal_packages/keybase/lib/pgp-key-store.cjsx
```diff
@@ -0,0 +1,391 @@
...
+                """
+
+          post = """
+                 </div>
+                 """
+          # can't use _.template :(
+          body = message.body.slice(0, blockStart) + pre + plaintext + post + message.body.slice(blockEnd)
+
+          # TODO if message is already in the cache, consider updating its TTL
+          timeout = 1000 * 60 * 30 # 30 minutes in ms
```
bengotow: Is there a scenario where the message would change after 30m? In most of N1 we assumes message objects are immutable if they aren't drafts.

internal_packages/keybase/lib/pgp-key-store.cjsx
```diff
@@ -0,0 +1,391 @@
...
+          @_msgStatus.push({"clientId": message.clientId, "time": Date.now(), "message": "Message decrypted with key #{keyprint}!"})
+          # re-render messages
+          MessageBodyProcessor.resetCache()
+          @trigger(@)
+        else
+          console.warn "Unable to decrypt message."
+          @_msgStatus.push({"clientId": message.clientId, "time": Date.now(), "message": "Unable to decrypt message."})
+
+  _onDraftChanged: (changes) ->
+    # every time the draft changes, get keys for all the recipients
```
bengotow: I think this funtionality is also implemented using that `setInterval` in the React component? Would be nice to keep triggering this only in the composer component so that we don't have to worry about accidentally loading keys for drafts the user doesn't care about or isn't editing! (From here, we can't tell the difference)

internal_packages/keybase/spec/encrypt-button-spec.cjsx
```diff
@@ -0,0 +1,143 @@
+{React, ReactDOM, ReactTestUtils, DraftStore, Message} = require 'nylas-exports'
+pgp = require 'kbpgp'
+
+EncryptMessageButton = require '../lib/encrypt-button'
+PGPKeyStore = require '../lib/pgp-key-store'
+
+describe "EncryptMessageButton", ->
+  beforeEach ->
+    key = """-----BEGIN PGP PRIVATE KEY BLOCK-----
```
bengotow: It'd be cool to move these test keys to a `fixtures` folder inside the spec folder for the plugin, and then load them in the specs via `key = fs.readFileSync('./fixtures/text-private-key.txt')`

internal_packages/keybase/lib/preferences-keybase.cjsx
```diff
@@ -0,0 +1,242 @@
...
+    </select>
+
+  _renderName: ->
+    <div>
+      <section className="public-key-picker">
+        Public Key: {@_renderPublicKeyPicker()}
+        <button className="btn key-name-btn" title="New key" onClick={@_saveNewPubKey}>New</button>
+        <button className="btn key-name-btn" title="Delete" onClick={@_deletePubKey}>Delete</button>
+        <button className="btn key-name-btn" title="Add" onClick={@_addOtherAddress}>Add</button>
+        <button className="btn key-name-btn" title="Add" onClick={@_removeOtherAddress}>Remove</button>
```
bengotow: Title should be Remove here?

internal_packages/keybase/lib/preferences-keybase.cjsx
```diff
@@ -0,0 +1,242 @@
...
+
+  _renderPrivateKeyPicker: ->
+    options = @state.privKeys.map (privKey) ->
+      <option value={privKey.addresses[0]} key={privKey.addresses[0]}>{privKey.addresses[0]}</option>
+
+    <select value={@state.selectedPrivKey?.addresses[0]} onChange={@_onSelectPrivKey}>
+      {options}
+    </select>
+
+  _renderName: ->
```
bengotow: It'd be nice to move this to another child React component, maybe called "MyKeyManager"? Maybe we could make MyKeyManager and FriendsKeyManager? haha I dunno.

internal_packages/keybase/stylesheets/main.less
```diff
@@ -0,0 +1,88 @@
+@import "ui-variables";
+@import "ui-mixins";
+
+.keybase-profile {
+  border: 1px solid black;
```
bengotow: Technically, we'll want to pull the appropriate variables for padding, background and border colors from `ui-variables` before we ship this. For example this would probably `border-color-primary`. Can wait for design pass though :-)

internal_packages/keybase/lib/pgp-key-store.cjsx
```diff
@@ -0,0 +1,401 @@
...
+      'Delete this key?',
+      'The key will be permanently deleted.',
+      ['Delete', 'Cancel']
+    )
+      fs.unlink(key.path, (err) =>
+        if (err)
+          @_displayError(err)
+      )
+
+  addAddressToKey: (profile, address) =>
```
DakotaNelson: TODO: calling this function will cause a duplicate key to be added (new file created and subsequently cached, old key not removed from the cache) that makes React very upset due to duplicate `key` props when it goes to render.

internal_packages/keybase/lib/decrypt-button.cjsx
DakotaNelson: The weird thing is that we don't actually decrypt the message here - just load the key(s) - and decryption is handled in the message preprocessor.

internal_packages/keybase/lib/pgp-key-store.cjsx
DakotaNelson: I think the way @logandavis wrote this makes sense - it seems to me like it's just a way to pass a boolean while also documenting what it's for instead of just having `somefunc(true)` in which case who knows what that true does.

Correct me if I'm wrong, Logan, but it makes sense to me.

internal_packages/keybase/lib/pgp-key-store.cjsx
DakotaNelson: The timeout is to ensure that decrypted messages don't hang around forever - so that you don't have to close and re-open N1 to re-encrypt your messages and ensure they're "safe" (ish, the decrypted version is still in memory, although we should perhaps change that). Same goes for private keys.

internal_packages/keybase/lib/pgp-key-store.cjsx
DakotaNelson: This is actually no longer needed at all - I took it right out.

internal_packages/keybase/lib/preferences-keybase.cjsx
DakotaNelson: There used to be a `KeyManager` and `KeybaseSearch` component, but the problem was that there's some bleedover - so, for instance, searching keybase could turn up someone you have a key saved for, and it was unclear how to handle that in two entirely different components.

internal_packages/keybase/lib/preferences-keybase.cjsx
```diff
@@ -0,0 +1,238 @@
...
+    PGPKeyStore.saveNewKey("benbitdiddle@icloud.com", TEST_KEY, isPub = true)
+
+  _saveNewPrivKey: =>
+    PGPKeyStore.saveNewKey("benbitdiddle@icloud.com", TEST_KEY, isPub = false)
+
+  _addOtherAddress: =>
+    PGPKeyStore.addAddressToKey(@state.pubKeys[0], "logan@nylas.com")
+
+  _removeOtherAddress: =>
+    PGPKeyStore.removeAddressFromKey(@state.pubKeys[0], "logan@nylas.com")
```
DakotaNelson: TODO add `ParticipantsTextField` to the key add/remove/etc. components Logan set up and remove these hardcoded values.

internal_packages/keybase/lib/preferences-keybase.cjsx
DakotaNelson: I take that back - they still both exist and are used. This is now vestigial and unused since Logan built `key-adder.cjsx` - I just removed it.

internal_packages/keybase/lib/pgp-key-store.cjsx
logandavis: Yes, that's why I'm doing that, and it's always seemed to work - although I admit I'm much more familiar with Python than CoffeeScript. I think the actually correct way to do this is with the expanded calls that assign inputs as object attributes - so instead of `isPub=true`, this'd be `{isPub: true}`, but then you gotta be careful to change it everywhere.

internal_packages/keybase/stylesheets/main.less
DakotaNelson: Yeah, this is definitely a placeholder until we put something not-terrible in place. :)

logandavis:
Merged in https://github.com/nylas/N1/commit/8f89ed79cbb97b0a917e883d3952507db6697e06

DakotaNelson:
:tada: :balloon: :tada:

PR #2016, opened 2016-04-21
Parse Office 365 names

pamdirac:
Office 365 likes to format email addresses as follows:
Last Name, First Name (Some Description) <email@address.org>

This causes Contact.firstName() (and hence sometimes displayName()) to return
"Last Name," which looks a bit odd in the message/thread views.

The intent of the commit is to correctly parse these names so that (using the
above example):
- firstName = "First Name"
- lastName = "Last Name (Some Description)"
- fullName = "First Name Last Name (Some Description)"

These behavioral changes only impact names containing a ','.

I don't know that this really provides exhaustive coverage of Office 365, and
keeping the description as part of the last name is not completely guilt free,
but it's not any worse than the previous state of affairs which also has
the description in the same field but combined instead with (misplaced)
first name data.

### DIFF
--- a/internal_packages/message-list/lib/message-participants.cjsx
+++ b/internal_packages/message-list/lib/message-participants.cjsx
@@ -58,7 +58,7 @@ class MessageParticipants extends React.Component
       if c.name?.length > 0 and c.name isnt c.email
         <div key={"#{c.email}-#{i}"} className="participant selectable">
           <div className="participant-primary" onClick={@_selectText}>
-            {c.name}
+            {c.fullName()}
           </div>
           <div className="participant-secondary">
             {"<"}<span onClick={@_selectText}>{c.email}</span>{">#{comma}"}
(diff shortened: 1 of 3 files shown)


bengotow:
Hi @pamdirac — sorry for the delay! This looks great. We've been wondering what to do with these sort of names for a while and I think this is a good solution. I feel like there might be cases where the `,` doesn't delimit <last name> and <first name>, but I'm not sure offhand. I think we should land this and then we'll see if anyone brings up counter-examples.

Thanks—we're actually shipping a release this week, so this will be live in a few days.

pamdirac:
Cool.  Agreed this might not be the end, but I am enjoying this improvement
locally, and I appreciate the merge.

On Wed, Apr 27, 2016 at 1:26 PM, Ben Gotow <notifications@github.com> wrote:

> Merged #2016 <https://github.com/nylas/N1/pull/2016>.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/nylas/N1/pull/2016#event-643219043>
>



-- 
John McNair
john@mcnair.org

PR #2017, opened 2016-04-21
Use long names

pamdirac:
Added an option to allow the user to see full contact names on message views.


pamdirac:
This is actually 2 commits based on top of parse-o365-names.  I'm running both together locally and made the changes at the same time.  These 2 commits could be merged independently if so desired.

Issue #1128, opened 2016-01-24
Horizontal scrollbar on wide messages is not visible


Acidter:
No horizontal scrolling letters. Or make zooming letter that it contains in the right window. This is a hindrance.

grinich:
I really can't understand what you're saying here.

Acidter:
A letter can only scroll vertically, horizontally to scroll can not be.
![1](https://cloud.githubusercontent.com/assets/1911400/12546144/d027588e-c358-11e5-8cd3-7fc116266800.jpg)

micphi:
Can confirm. 

If I'm using two panel layout and the right panel isn't wide enough to display the entirety of an email's content, there is no horizontal scrollbar to allow for easy scrolling without having to resort to left/right arrow keys or window resizing.

bengotow:
Hey folks—I believe on devices with trackpads, you can pan on the trackpad to scroll the message left and right, but there's no affordance for this. I'll see if we can get it addressed in the future.

bengotow:
Merging in #785

calope:
I can reproduce the problem in the Windows version 0.4.19.
The problem is with wide images. It's reproducible both single panel as two panel Layout

stanfieldr:
As a temporary solution, you guys can click somewhere on the email and use your left and right arrow keys to scroll horizontally if you don't have a trackpad.

jackiehluo:
Hey! A fix is shipping now in 0.4.40, which will be hitting the auto-updaters soon. (If you want, you can grab the update now on the Releases page.) Let us know if you're still experiencing the issue after trying it! Thanks!

PR #2028, opened 2016-04-22
horizontal scrollbar does not appear when viewing wide email

stanfieldr:
This fixes issue #1128 

The horizontal scrollbar was not showing up in two cases for me:
1) A badly formatted email had some weird span with white-space: nowrap which caused the words to run off the screen
2) A wide image was only showing the left part.

Either case, the problem is the iframe not scrolling horizontally. I fiddled with the iframe for a while and couldn't get it to scroll, so I ended up setting overflow on the body of the iframe page instead.

**Before:**
selection

**After (notice border to the right):**
selection

### DIFF
--- a/internal_packages/message-list/lib/email-frame.jsx
+++ b/internal_packages/message-list/lib/email-frame.jsx
@@ -82,13 +82,18 @@ export default class EmailFrame extends React.Component {
   }
 
   _getFrameHeight = (doc) => {
+    let height = 0;
+
     if (doc && doc.body) {
-      return doc.body.scrollHeight;
+      height = doc.body.scrollHeight;
     }
+
     if (doc && doc.documentElement) {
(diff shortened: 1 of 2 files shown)


stanfieldr:
Now that I thought about this a bit more, it might be more convenient for the user if the horizontal scroll was for the whole thread instead of a particular message. The message width would be expanded using `iframe.scrollWidth` so all text is visible in the iframe. And the frame with the max width in the thread would be the width of the white area of every message.

By having the scrollbar on the whole thread, it would make it easier to scroll long messages horizontally without needing to scroll to the bottom.

I'm not really familiar with the code base though, so I would probably have to look around a bit more.

Anyway, this PR at least fixes the functionality aspect.

bengotow:
Hey this looks great! Sorry for the delay—We'll include this in the next release.

PR #2041, opened 2016-04-23
Update title for system tray

akshar-patel:
Using only "Menu Bar" confuses Linux users because ''Menu Bar'' word is only used withing Mac OS. So, also use "System Tray" word.


bengotow:
Looks great, thanks!

PR #2168, opened 2016-05-08
Added Less Is More

P0WW0W:
Just included ui-less-is-more to internal-packages, and edited the README.md to reflect updates.


internal_packages/ui-less-is-more/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "less-is-more",
```
bengotow: Can we change this to `ui-less-is-more`?

internal_packages/ui-less-is-more/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+[![Message](./images/composer-window.png)](./images/composer-window.png)
+
+[More images can be found at →](http://www.alexanderadkins.com/projects/less-is-more)
+
+## Installation
+
+[Download Less Is More](https://github.com/P0WW0W/less-is-more/archive/master.zip)
+
+##### Step 1
```
bengotow: Can you drop the installation steps and screenshot, since folks can use N1's theme picker to try it out / select it?

bengotow:
Looks great! I just added a few comments but we should definitely get this landed for the next release. 👍

PR #2343, opened 2016-05-29
warns suspicious message only on differing domain

### LINKED ISSUE #2342: Reply-to differing but on same domain should not be considered suspicious
##### Are there any related issues?
No.

OS X 10.11.5

0.4.40

--

**Bug:**

The warning:
> This message looks suspicious!
> It originates from X but replies will go to Y.

need not occur when, for example `X = foo@example.org` and `Y = bar@example.org`.

A common and legitimate use of modifying the `Reply-to` header is in mailing list management, those that aren't used as a sort of dicussion thread, but rather for information coming from a single source such as a club or society; replies may most commonly be sent to the organisers, which requires a different reply address.

Of course, this use case may well occur occasionally across different domains, but I think that should be niche enough to let a warning fly (perhaps in the more distant future, there could be an option to "ignore for this sender/reply-to combination").

No.

I assume not.

##### Is the issue reproducible with a particular attachment, message, signature, etc?
N/A.

OJFord:
For example, the following is *not* suspicious:
```
From: foo@example.org
Reply-To: bar@example.org
```

While the following remains suspicious:
```
From: foo@example.org
Reply-To: foo@example.com
``` 

This commit fixes #2342.

### DIFF
--- a/internal_packages/phishing-detection/lib/main.jsx
+++ b/internal_packages/phishing-detection/lib/main.jsx
@@ -48,9 +48,11 @@ class PhishingIndicator extends React.Component {
     // phishing attempt boils down to checking the `replyTo` attributes on
     // `Message` models from `MessageStore`.
     if (message && message.replyTo && message.replyTo.length !== 0) {
-      const from = message.from[0].email;
-      const replyTo = message.replyTo[0].email;
-      if (replyTo !== from) {
+      const from = message.from[0].email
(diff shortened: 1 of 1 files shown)


bengotow:
Looks goo! Can you sign the CLA and we'll get this merged?

OJFord:
Cheers, done.

PR #2420, opened 2016-06-12
Don't show phishing warning if Reply-To and From are on the same domain; even with differing subdomain

OJFord:
This commit improves on nylas/N1#2343 (which allowed From and
Reply-To to be different aliases @ the same domain) by allowing the
address to differ by a subdomain.

For example, the pairs:

    From:     foo@bar.com
    Reply-To: foo@sub.bar.com

    From:     foo@bar.co.uk
    Reply-To: foo@other.bar.co.uk

    From:     foo@bar.com
    Reply-To: foobar@bar.com
are each now allowed without warning.

### DIFF
--- a/internal_packages/phishing-detection/lib/main.jsx
+++ b/internal_packages/phishing-detection/lib/main.jsx
@@ -7,6 +7,8 @@ import {
   MessageStore,
 } from 'nylas-exports';
 
+const tld = require('tld');
+
 // Notice that this file is `main.cjsx` rather than `main.coffee`. We use the
 // `.cjsx` filetype because we use the CJSX DSL to describe markup for React to
 // render. Without the CJSX, we could just name this file `main.coffee` instead.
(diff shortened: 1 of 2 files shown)


OJFord:
(CI failed due to a syntax error, fixed up to maintain cherry-pick etc.)

bengotow:
Hey—thanks for reporting this! I'm going to go ahead and merge this, but we may need to look at significantly changing the way the phishing detection plugin works, since a spammer could be emailing you "from" one gmail.com account with their reply-to set to another gmail.com account. We're hoping to expose the DKIM headers and a few other things on our backend, and that would allow us to make this a lot more sophisticated!

OJFord:
No worries, thanks for merging :)

And agreed - at the moment it's name makes it sound far more sophisticated than it really is!

This has just been bugging me a lot, since the vast majority of my emails (i.e. all communications from companies, shipping notifications, etc. and in particular anything that uses some sort of 'email campaign manager' tool) report a suspicious reply-to.

False-positives aren't even merely annoying - they're also dangerous to some extent, since they train us to ignore the warning when it may be real.

PR #2485, opened 2016-06-19
Make display of onboarding page conditional.

### LINKED ISSUE #2468: With new N1, can't connect to sync engine
After upgrading N1 to the new 0.45 release, I can no longer connect to my locally installed sync-engine. Launching N1, I get a dialog to sign-up for a Nylas ID, and then an error that N1 can't connect to an onboarding page at localhost:5009.  I've tried updating the sync-engine, but no luck.  

Also, it continues to be impossible to change a password if you're using the local sync engine. You end up creating an entirely new account entry.  Besides re-downloading everything, that also means there are then multiple account entries in the database with the same name, and API calls begin to fail for that reason. 

Ubuntu 16.04

.45

--

**Bug?**
No

No

jonasws:
When running against a custom Sync Engine, viewing the onboarding page
on startup does not make much sense. This here adds code that checks for
the `env` of the Nylas config and skips the onboarding if it is `local`
or `custom`.

This fixes #2468.

### DIFF
--- a/src/browser/application.es6
+++ b/src/browser/application.es6
@@ -193,8 +193,11 @@ export default class Application extends EventEmitter {
     const accounts = this.config.get('nylas.accounts');
     const hasAccount = accounts && accounts.length > 0;
     const hasN1ID = this.config.get('nylas.identity.id');
+    const env = this.config.get('env');
+    const isLocalOrCustom = (env === 'local' || env === 'custom');
 
-    if (hasAccount && hasN1ID) {
+
+    if (isLocalOrCustom || (hasAccount && hasN1ID)) {
       this.windowManager.ensureWindow(WindowManager.MAIN_WINDOW);
       this.windowManager.ensureWindow(WindowManager.WORK_WINDOW);
     } else {


systemeintegratedsolutions:
Has this been released to production then? (sorry I'm still getting used to how this Github process works)

jonasws:
No, the pull request has not been merged.

OliverCole:
How do we get someone to merge it? :)

bengotow:
Hey folks! Has anyone tried to run the app with this patch applied? I think it'll still be necessary to populate the `config.json` file with an identity and put an identity token into the user's system keychain. Skipping the dialog is probably not enough!

Otherwise when the app runs, exceptions will be thrown as you move through the interface (because the UI queries for the user identity) and also when N1 tries to make any API request (because of safety checks like this: https://github.com/nylas/N1/blob/stable/src/flux/nylas-api-request.es6#L37).

jonasws:
@bengotow I have used this patch since I submitted it and it works fine. I did try to analyze the code in order to see if there were other places in the code that were relying on the N1 ID feature, but I couldn't find any at the time.

bengotow:
Cool! Can you double check that your previous credentials aren't saved in the keychain or config.json? It might be reading you old access token and config and continuing to work. (if you logged into N1 before creating this patch). On a Mac you can launch Keychain Access and the N1 token should appear in the "All Items" list. There are entries for each of your email accounts and also one that says "Account: Nylas Account" which is your N1 Identity token:

image

jonasws:
I dobule chceked, nd all that is there (in my GNOME keyring) are the tokens put there were for my email accounts. None for the Nylas ID account. So all should be good!

jonasws:
@bengotow any plan on when to merge this? Seems it would help a lot of people. At least people are asking questions about this on Slack.

alexanderadam:
I can confirm that this change works (under current Ubuntu with KDE)

FranckKe:
Works fine for me too (Manjaro/xfce). Thank you for the PR :+1:

bengotow:
Hey folks—thanks for the feedback. I'm going to go ahead and merge this, but we're working on a broader patch for users who are hosting the sync engine themselves. We never intended for `IdentityStore.identity()` to return `null` (as it does in this patch), so you may run in to small issues with things like the preferences UI. Stay tuned!

PR #2514, opened 2016-06-25
Mention test focusing and skipping in CONTRIBUTING

courajs:


courajs:
I include this specifically here because the jasmine 1.3 docs don't mention `fdescribe` or `fit`

bengotow:
Thanks—this is definitely helpful!

PR #2517, opened 2016-06-26
Change pending unread notification stack to be a queue

courajs:
Relatively minor, but I noticed it while poking around in the file, and
it started to bother me.

Consider the case where a user launches N1, and has 3 unreads since
their last launch. Additionally, after N1 has launched, they begin
receiving a new mail every 2 seconds.

Current behavior would be a notification for 1 of the 3 unread emails,
then every 2 seconds a notification for the most recently received mail.
They would only finally receive notifications for the other 2 original
messages once they stopped getting new mail.


bengotow:
Hi @courajs — thanks for submitting this. Looks great! Definitely better to have different subject lines for each email in those tests. Thanks—

PR #2534, opened 2016-06-30
pgp-plugin update

logandavis:
This update to the PGP plugin overhauls the Decrypt UI, adds a bunch of error messaging, and corrects some bugs in the encryption key picker modal.

### DIFF
--- a/internal_packages/keybase/lib/decrypt-button.cjsx
+++ b/internal_packages/keybase/lib/decrypt-button.cjsx
@@ -3,6 +3,7 @@ PGPKeyStore = require './pgp-key-store'
 {remote} = require 'electron'
 PassphrasePopover = require './passphrase-popover'
 pgp = require 'kbpgp'
+_ = require 'underscore'
 
 class DecryptMessageButton extends React.Component
 
@@ -46,7 +47,7 @@ class DecryptMessageButton extends React.Component
     else
       popoverTarget = event.target.getBoundingClientRect()
       Actions.openPopover(
-        <PassphrasePopover onPopoverDone={ @_decryptPopoverDone } />,
+        <PassphrasePopover addresses={_.pluck(message.to, "email")} onPopoverDone={ @decryptPopoverDone } />,
         {originRect: popoverTarget, direction: 'down'}
       )
 
@@ -65,11 +66,11 @@ class DecryptMessageButton extends React.Component
     else
       popoverTarget = event.target.getBoundingClientRect()
       Actions.openPopover(
-        <PassphrasePopover onPopoverDone={ @_decryptAttachmentsPopoverDone } />,
+        <PassphrasePopover addresses={_.pluck(message.to, "email")} onPopoverDone={ @decryptAttachmentsPopoverDone } />,
         {originRect: popoverTarget, direction: 'down'}
       )
 
-  _decryptPopoverDone: (passphrase) =>
+  decryptPopoverDone: (passphrase) =>
     {message} = @props
     for recipient in message.to
       # right now, just try to unlock all possible keys
@@ -78,7 +79,7 @@ class DecryptMessageButton extends React.Component
       for privateKey in privateKeys
         PGPKeyStore.getKeyContents(key: privateKey, passphrase: passphrase)
 
-  _decryptAttachmentsPopoverDone: (passphrase) =>
+  decryptAttachmentsPopoverDone: (passphrase) =>
     {message} = @props
     for recipient in message.to
       privateKeys = PGPKeyStore.privKeys(address: recipient.email, timed: false)
--- a/internal_packages/keybase/lib/decrypt-button.cjsx
+++ b/internal_packages/keybase/lib/decrypt-button.cjsx
@@ -32,18 +32,17 @@ class DecryptMessageButton extends React.Component
     @unlistenKeystore()
 
   _onKeystoreChange: ->
-    @setState(@_getStateFromStores())
     # every time a new key gets unlocked/fetched, try to decrypt this message
     if not @state.isDecrypted
       PGPKeyStore.decrypt(@props.message)
+    @setState(@_getStateFromStores())
 
   _onClickDecrypt: (event) =>
-    {message} = @props
     popoverTarget = event.target.getBoundingClientRect()
     if @_noPrivateKeys()
       Actions.openPopover(
         <PrivateKeyPopover
-          addresses={_.pluck(message.to, "email")}
+          addresses={_.pluck(@props.message.to, "email")}
           callback={() => @_openPassphrasePopover(popoverTarget, @decryptPopoverDone)}/>,
         {originRect: popoverTarget, direction: 'down'}
       )
@@ -55,12 +54,11 @@ class DecryptMessageButton extends React.Component
     dialog.showErrorBox('Decryption Error', err.toString())
 
   _onClickDecryptAttachments: (event) =>
-    {message} = @props
     popoverTarget = event.target.getBoundingClientRect()
     if @_noPrivateKeys()
       Actions.openPopover(
         <PrivateKeyPopover
-          addresses={_.pluck(message.to, "email")}
+          addresses={_.pluck(@props.message.to, "email")}
           callback={() => @_openPassphrasePopover(popoverTarget, @decryptAttachmentsPopoverDone)}/>,
         {originRect: popoverTarget, direction: 'down'}
       )
@@ -68,8 +66,7 @@ class DecryptMessageButton extends React.Component
       @_openPassphrasePopover(popoverTarget, @decryptAttachmentsPopoverDone)
 
   decryptPopoverDone: (passphrase) =>
-    {message} = @props
-    for recipient in message.to
+    for recipient in @props.message.to
       # right now, just try to unlock all possible keys
       # (many will fail - TODO?)
       privateKeys = PGPKeyStore.privKeys(address: recipient.email, timed: false)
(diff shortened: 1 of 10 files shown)


internal_packages/keybase/lib/decrypt-button.cjsx
```diff
@@ -29,80 +32,112 @@ class DecryptMessageButton extends React.Component
...
     decryptAttachments = false
-    if @state.encryptedAttachments?.length == 1
-      decryptAttachments = <button onClick={ @_onClickDecryptAttachments } className="btn btn-toolbar">Decrypt Attachment</button>
-      title = "Attachment Encrypted"
-    else if @state.encryptedAttachments?.length > 1
-      decryptAttachments = <button onClick={ @_onClickDecryptAttachments } className="btn btn-toolbar">Decrypt Attachments</button>
-      title = "Attachments Encrypted"
-
+    if @state.encryptedAttachments?.length >= 1
+      title = if @state.encryptedAttachments?.length == 1 then "Attachment Encrypted" else "Attachments Encrypted"
```
bengotow: Since this is inside the ?.length > 1 block, we could safely omit the ? here

internal_packages/keybase/lib/decrypt-button.cjsx
```diff
@@ -29,80 +32,112 @@ class DecryptMessageButton extends React.Component
...
+      buttonLabel = if @state.encryptedAttachments?.length == 1 then "Decrypt Attachment" else "Decrypt Attachments"
```
bengotow: This will look great once we're able to convert this to ES6 ;-) and get some ternaries going on.

internal_packages/keybase/lib/decrypt-button.cjsx
```diff
@@ -29,80 +32,112 @@ class DecryptMessageButton extends React.Component
...
       return false
 
-    # TODO a message saying "this was decrypted with the key for ___@___.com"
-    title = if @state.isDecrypted then "Message Decrypted" else "Message Encrypted"
+    title = "Message Encrypted"
+    decryptLabel = "Decrypt"
+    borderClass = "border"
+    decryptClass = "decrypt-bar"
+    if @state.status?
+      console.warn(@state.status)
```
bengotow: Can we remove this?

internal_packages/keybase/lib/decrypt-button.cjsx
```diff
@@ -29,80 +32,112 @@ class DecryptMessageButton extends React.Component
...
+      if @state.status.indexOf("Message decrypted") >= 0
+        title = @state.status
+        borderClass = "border done-border"
+        decryptClass = "decrypt-bar done-decrypt-bar"
```
bengotow: It might make sense to boil this all down to a single variable, like `class=done`, and then say "border #{class}-border", "decrypt-bar #{class}-decrypt-bar", etc.

internal_packages/keybase/lib/key-manager.cjsx
```diff
@@ -38,7 +38,7 @@ class KeyManager extends React.Component
     popoverTarget = event.target.getBoundingClientRect()
 
     Actions.openPopover(
-      <PassphrasePopover identity={identity} onPopoverDone={ @_exportPopoverDone } />,
+      <PassphrasePopover identity={identity} addresses={identity.addresses} onPopoverDone={ @_exportPopoverDone } />,
```
bengotow: [Is it necessary to pass these separately if they're inside identity already? Honest question, might make sense given the popover's use elsewhere.]

internal_packages/keybase/lib/modal-key-recommender.cjsx
```diff
@@ -68,8 +68,23 @@ class ModalKeyRecommender extends React.Component
...
       )
         return
-    Actions.closePopover()
-    @props.callback(@state.identities)
+
+    if _.every(@state.identities, (identity) -> identity.key?)
+      Actions.closePopover()
+      @props.callback(@state.identities)
+    else
+      emptyIdents = _.filter(@state.identities, (identity) -> !identity.key?)
```
bengotow: Might make things slightly cleaner to bump this up a line and then say `if emptyIdents.length === 0` instead of _.every!

internal_packages/keybase/lib/modal-key-recommender.cjsx
```diff
@@ -68,8 +68,23 @@ class ModalKeyRecommender extends React.Component
...
+      newIdents = []
+      for idIndex of emptyIdents
+        identity = emptyIdents[idIndex]
+        if idIndex < emptyIdents.length - 1
+          PGPKeyStore.getKeyContents(key: identity, callback: (identity) => newIdents.push(identity))
+        else
+          PGPKeyStore.getKeyContents(key: identity, callback: (identity) =>
+            newIdents.push(identity)
+            @props.callback(newIdents)
```
bengotow: Hmm, I'm not sure about the intended behavior here. It looks like this will call props.callback three times if there are three empty idents, once with a [A], again with [A, B], and then again with [A,B,C]. Maybe we should use an `async` library method to collect all of the results from `getKeyContents` before firing the callback with all of them?

internal_packages/keybase/lib/passphrase-popover.cjsx
```diff
@@ -1,25 +1,67 @@
...
+pgp = require 'kbpgp'
 
 module.exports =
 class PassphrasePopover extends React.Component
   constructor: ->
-    @state = {passphrase: ""}
+    @state = {
+      passphrase: ""
+      placeholder: "PGP private key password"
+      classNames: "key-passphrase-input form-control"
```
bengotow: Typically  you want React component state to reflect the conceptual state of the component, rather than what that means in terms of HTML text / classes / styles. I think this will work, but it'd be cleaner to say `@state = {error: false}` and then in the render method, add an extra className if `error = true`.

internal_packages/keybase/lib/passphrase-popover.cjsx
```diff
@@ -1,25 +1,67 @@
...
+      <button className="btn btn-toolbar" onClick={@_validatePassphrase}>Done</button>
     </div>
 
   _onPassphraseChange: (event) =>
     @setState
       passphrase: event.target.value
+      placeholder: "PGP private key password"
+      classNames: "key-passphrase-input form-control"
+
+  _onCarriageReturn: (event) =>
```
bengotow: Stupid little nit, but it'd be nice if this was `_onKeyUp` since it technically isn't only fired on carriage returns.

internal_packages/keybase/lib/passphrase-popover.cjsx
```diff
@@ -1,25 +1,67 @@
...
+            if err
+              console.warn err
+            else
+              km.unlock_pgp { passphrase: passphrase }, (err) =>
+                if err
+                  if parseInt(keyIndex, 10) == privateKeys.length - 1
+                    if parseInt(emailIndex, 10) == @props.addresses.length - 1
+                      # every key has been tried, the password failed on all of them
+                      @setState
+                        passphrase: ""
```
bengotow: It may be necessary to keep an internal variable in this popover component to check that the component is still mounted before calling setState here. Essentially the user can call validatePassphrase, then close the popover, and by the time this line of code runs (asynchronously) the component is technically gone an setState will throw an exception.

Typically the way to guard against this is to add an internal var like this:

```
componentDidMount() {
  this._mounted = true;
}

componentWillUnmount() {
  this._mounted = false;
}

// in callback, before calling setState, check that _mounted is still true!

File: internal_packages/keybase/lib/pgp-key-store.cjsx [hunk 11]
```diff
@@ -155,11 +155,14 @@ class PGPKeyStore extends NylasStore
...
     else
       fingerprint = identity.fingerprint()
-      kb.getUser(fingerprint, 'key_fingerprint', (err, user) =>
-        if user?.length == 1
-          identity.keybase_profile = user[0]
+      if fingerprint?
+        kb.getUser(fingerprint, 'key_fingerprint', (err, user) =>
+          if err
+            console.error(err)
+          if user?.length == 1
```
bengotow: Really weird that the keybase API returns an array called user...

internal_packages/keybase/lib/private-key-popover.cjsx
```diff
@@ -0,0 +1,138 @@
...
+    saveBtnClass = if !(@state.validKeyBody) then "btn modal-done-button btn-disabled" else "btn modal-done-button"
+    saveButton = <button className={saveBtnClass} disabled={!(@state.validKeyBody)} onClick={@_onDone}>Save</button>
+
+    <div className="private-key-popover" tabIndex=0>
+      <span key="title" className="picker-title"><b>No PGP private key found.<br/>Add a key for {@_renderAddresses()}</b></span>
+      <div className="key-add-buttons">
+        <button className="btn btn-toolbar paste-btn" onClick={@_onClickPaste}>Paste in a Key</button>
+        <button className="btn btn-toolbar import-btn" onClick={@_onClickImport}>Import from File</button>
+      </div>
+      {if (@state.import or @state.paste) and !@state.validKeyBody and @state.keyBody != "" then errorBar}
```
bengotow: This totally works, but it can get a little confusing to see what the final state will be with so many conditional elements. If this were to become more complicated, I think we'd want to consider:

1. Setting a CSS class on the container, and then using CSS display:none to hide elements which aren't relevant, rather than making all the rendering conditional.

2. Identifying commonality between different states (maybe `(@state.import or @state.paste)`) and creating two render methods. Even though this might cause some duplication of elements that are present in both cases, find replace is easy and following conditionals is hard - just having renderImport / renderNonimport might be easier to maintain.

internal_packages/keybase/lib/private-key-popover.cjsx
```diff
@@ -0,0 +1,138 @@
...
+
+  _renderAddresses: =>
+    signedIn = _.pluck(AccountStore.accounts(), "emailAddress")
+    suggestions = _.intersection(signedIn, @props.addresses)
+
+    if suggestions.length == 1
+      addresses = <span>{suggestions[0]}.</span>
+    else if suggestions.length > 1
+      options = suggestions.map((address) => <option value={suggestions.indexOf(address)} key={suggestions.indexOf(address)}>{address}</option>)
+      addresses =
```
bengotow: This `addresses` object is assigned to but doesn't seem to be used?

internal_packages/keybase/lib/private-key-popover.cjsx
```diff
@@ -0,0 +1,138 @@
...
+        <select value={@state.selectedAddress} onChange={@_onSelectAddress} style={{minWidth: 150}}>
+          {options}
+        </select>
+    else
+      console.error("How did you receive a message that you're not in the TO field for?")
```
bengotow: In cases like this, we may want to actually `throw new Error("How did you..."`. That way, this will get reported to our backend!

internal_packages/keybase/lib/decrypt-button.cjsx
logandavis: Might be useful in case of new and different error messages, but sure, I've removed it.

internal_packages/keybase/lib/key-manager.cjsx
logandavis: Yeah, over in decrypt-button the popover gets passed the To: field as addresses.

internal_packages/keybase/lib/modal-key-recommender.cjsx
logandavis: Done.

internal_packages/keybase/lib/passphrase-popover.cjsx
logandavis: Done.

internal_packages/keybase/lib/pgp-key-store.cjsx
logandavis: And that it needs to be one user long. Are people sharing public keys? 😞

internal_packages/keybase/lib/private-key-popover.cjsx
logandavis: Hmm? It gets used up in _render when _renderAddresses() is called. That dropdown definitely shows up...

internal_packages/keybase/lib/private-key-popover.cjsx
logandavis: Done.

internal_packages/keybase/lib/passphrase-popover.cjsx
logandavis: Done.

internal_packages/keybase/lib/passphrase-popover.cjsx
logandavis: Done.

internal_packages/keybase/lib/decrypt-button.cjsx
logandavis: Done.

internal_packages/keybase/lib/modal-key-recommender.cjsx
```diff
@@ -68,8 +68,23 @@ class ModalKeyRecommender extends React.Component
         ['Encrypt', 'Cancel']
       )
         return
-    Actions.closePopover()
-    @props.callback(@state.identities)
+
+    emptyIdents = _.filter(@state.identities, (identity) -> !identity.key?)
```
YurySolovyov: Why not use native `Array.prototype.filter`?

PR #2554, opened 2016-07-05
Fix failing clean script

carsonip:
The clean script fails since productName is "Nylas N1" which contains a space and it appears to pkill 2 patterns. This is causing the script to stop in my linux machine. Now the productName is wrapped with quotes to prevent this from happening.


bengotow:
Thanks—this looks great!

PR #2600, opened 2016-07-12
Fix incorrect port number when using https

carsonip:
When APIRoot is using https, the port will forced to be 443 even when APIRoot
indicates another port. This is because of the line options.port = 443
overwriting the correct options.port extracted by url.parse. This is fixed by
removing the problematic line. The default port is still 443 if you look into
https.js. In case we want a different default port in the future, we can do
options.port = options.port || 443.


bengotow:
Thanks @carsonip — this looks great.

PR #2622, opened 2016-07-14
24hr

jathpala:
Addresses issue #682 

Adds support for displaying times on a 24hr clock. I think this allows display of 24hr time in all places where time can be displayed. The option to use 24hr time is selected in preferences.

As part of this update I have removed the old date/time functions in flux/models/utils and created new versions of these functions in date-utils. All time display should now be based on date-utils.

The only major flaw I know of in this implementation is when toggling between 12 and 24 hour time, Nylas needs to be refreshed before the change takes effect. If anyone knows how to automatically refresh the date/time display I could update this behaviour.

What do you think? This is my first patch for Nylas so sorry if my code structure is different to what is expected for this project. Happy to make changes as needed.

### DIFF
--- a/internal_packages/message-list/lib/message-timestamp.cjsx
+++ b/internal_packages/message-list/lib/message-timestamp.cjsx
@@ -15,9 +15,13 @@ class MessageTimestamp extends React.Component
     +nextProps.date isnt +@props.date or nextProps.isDetailed isnt @props.isDetailed
 
   render: =>
+    if @props.isDetailed
+      formattedDate = DateUtils.mediumTimeString(@props.date)
+    else
+      fromattedDate = DateUtils.shortTimeString(@props.date)
     <div className={@props.className}
          title={DateUtils.fullTimeString(@props.date)}
-         onClick={@props.onClick}>{DateUtils.shortTimeString(@props.date)}</div>
+         onClick={@props.onClick}>{formattedDate}</div>
 
   # Stubbable for testing. Returns a `moment`
   _today: -> moment.tz(Utils.timeZone)
--- a/internal_packages/message-list/lib/message-timestamp.cjsx
+++ b/internal_packages/message-list/lib/message-timestamp.cjsx
@@ -15,30 +15,9 @@ class MessageTimestamp extends React.Component
     +nextProps.date isnt +@props.date or nextProps.isDetailed isnt @props.isDetailed
 
   render: =>
-    msgDate = moment.tz(@props.date, Utils.timeZone)
-    nowDate = @_today()
-    formattedDate = @_formattedDate(msgDate, nowDate, @props.isDetailed)
     <div className={@props.className}
          title={DateUtils.fullTimeString(@props.date)}
-         onClick={@props.onClick}>{formattedDate}</div>
-
-  _formattedDate: (msgDate, now, isDetailed) =>
-    timeFormat = DateUtils.getTimeFormat upperCase: true
-    if isDetailed
-      return msgDate.format "MMMM D, YYYY [at] #{timeFormat}"
-    else
-      diff = now.diff(msgDate, 'days', true)
-      isSameDay = now.isSame(msgDate, 'days')
-      if diff < 1 and isSameDay
-        return msgDate.format timeFormat
-      if diff < 1.5 and not isSameDay
-        timeAgo = msgDate.from now
-        monthAndDay = msgDate.format timeFormat
-        return monthAndDay + " (" + timeAgo + ")"
-      if diff >= 1.5 and diff < 365
-        return msgDate.format "MMM D"
-      if diff >= 365
-        return msgDate.format "MMM D, YYYY"
+         onClick={@props.onClick}>{DateUtils.shortTimeString(@props.date)}</div>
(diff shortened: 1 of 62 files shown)


bengotow:
Hey! Thanks for contributing this—overall it looks good and it's something we're definitely interested in getting into the next release.

internal_packages/thread-list/lib/thread-list-columns.cjsx
```diff
@@ -17,11 +18,14 @@ ThreadListParticipants = require './thread-list-participants'
 ThreadListStore = require './thread-list-store'
 ThreadListIcon = require './thread-list-icon'
 
+{### Get and format either last sent or last received timestamp depending on thread-list being viewed ###}
```
bengotow: I think the wrapping `{` and `}` here may be an error?

internal_packages/message-list/lib/message-timestamp.cjsx
```diff
@@ -15,30 +15,9 @@ class MessageTimestamp extends React.Component
     +nextProps.date isnt +@props.date or nextProps.isDetailed isnt @props.isDetailed
 
   render: =>
-    msgDate = moment.tz(@props.date, Utils.timeZone)
-    nowDate = @_today()
-    formattedDate = @_formattedDate(msgDate, nowDate, @props.isDetailed)
     <div className={@props.className}
```
bengotow: It looks like we lost the isDetailed option here—we may need to check and see where / if that was in use anywhere.

src/date-utils.es6
```diff
@@ -156,21 +156,86 @@ const DateUtils = {
...
+   * @param {Date} datetime - Timestamp
+   * @return {String} Formated date/time
+   *
+   * The returned date/time format depends on how long ago the timestamp is.
+   */
+  shortTimeString(datetime) {
+    const diff = moment().diff(datetime, 'days', true)
+    let format = null
+
+    if (diff <= 1) {
```
bengotow: I think we actually want to pull the more advanced logic from the function above into here:

```
 -      isSameDay = now.isSame(msgDate, 'days')		
 -      if diff < 1 and isSameDay		
 -        return msgDate.format timeFormat		
 -      if diff < 1.5 and not isSameDay		
 -        timeAgo = msgDate.from now		
 -        monthAndDay = msgDate.format timeFormat		
 -        return monthAndDay + " (" + timeAgo + ")"		
```

Basically, this changes the behavior a bit:

- Email sent 0-24 hours ago: Just the time (eg: 9AM)
- Email sent 24-36 hours ago: Both the time and the date "July 17 (1AM)"
- Email sent 36+ hours ago: Just the date

I think having that intermediate state is important because people still want the time resolution. I believe it was added in a PR a while back!

src/date-utils.es6
```diff
@@ -156,21 +156,86 @@ const DateUtils = {
...
+
+
+  /**
+   * Return a long format date/time
+   *
+   * @param {Date} datetime - Timestamp
+   * @return {String} Formated date/time
+   */
+  fullTimeString(datetime) {
+    const tz = moment.tz.guess()
```
bengotow: I think that moment.tz.guess does a lot of work internally when the timezone isn't easy to determine. (It does a bunch of date math with common dates to identify what the time zone must be.) Since we call this function really often (as the user is scrolling in the thread list) and performance is super important, I'd pull this back out into global scope at the top of the file.

That /does/ mean that the user changing timezones won't be detected. We might want to create a getter method for it, and add a bit of logic to re-compute it at most every five minutes, or something like that if having it update is important. (Might want to make it a separate PR to avoid expanding the scope of this one too much.)

jathpala:
Thanks for the feedback. I have updated my code to incorporate these changes.

1. Fixed the comment. Thanks for that.

2. I've re-enable support for the isDetailed property. I've done this with a new function mediumTimeString() which displays date and time in a similar format to the original implementation.

3. shortTimeString now implements the more advance logic for finer grained display of dates and times. I've implemented it so it displays more or less as your description suggests it should (i.e "Jul 17, 1:30 am"). I think the code from the original function displayed it as time ago (i.e "Jul 17 (28 hours ago)"). I can change it back to time ago format if you like but I think simple date and time makes more sense.

4. Moved the call to moment.tz() to the global scope as suggested. I think implementing as a getter to allow updates to timezone would be relatively simple to do but I guess this would be a rare use case and maybe just better to leave a simpler implementation as we have now. If the user wants to update timezone, they can always just restart the program.

I still need to test this a bit more and I'll try to write some specs for it before it's ready to merge but would love any further feedback at this stage. Also, I haven't looked at issue #2652 and whether that is affected by these changes yet.

jathpala:
So having tested this a little bit now I think it works as expected. The main limitation with this code currently is that when changing time display to/from 24 hour the changes aren't applied until the app is refreshed. If anyone knows how to automatically trigger a refresh of the times displayed I'd like to try and add this in at some point. Otherwise hopefully this is ready to merge.

bengotow:
Looks great—thanks! We'll test this a bit more on master this week. Thanks for converting things to ES6 as well.

Issue #2410, opened 2016-06-10
paste kills unfinished addresses in composer


grinich:
### try this:

1. copy: **google.com**
2. go to the composer and type **joe@**
3. use CMD+V to paste

what do you get?
![image](https://cloud.githubusercontent.com/assets/92417/15953992/73e98328-2e86-11e6-857a-3b8ae9227e37.png) ![image](https://cloud.githubusercontent.com/assets/92417/15953993/770f0636-2e86-11e6-9fba-bb7113bbcbbf.png)

---

### This _should_ just append it so it looks like this: 

![image](https://cloud.githubusercontent.com/assets/92417/15953999/8772d688-2e86-11e6-8f61-7e40e6c2d277.png)

and then enter or tab tokenizes it properly 

![image](https://cloud.githubusercontent.com/assets/92417/15954011/b497c2c2-2e86-11e6-9d7d-b187bfd9a297.png)

jathpala:
I think i have come up with a fix for this (see pull request #2635).

This solves the problem more-or-less as suggested above. Pasted text is just appended to already existing text. If after pasting, the input field now looks like an email address, it will automatically be converted to a button like the last picture above (no need to press space, tab or enter). If after pasting, the input field does not look like a full address then it will remain as just text waiting for further input.

I believe this is the behaviour that most people would want and expect. There is the issue that if you wanted to send an email to 'test@email.com.au' but for some reason only copied 'email.com' to the clipboard before typing 'test@' then hitting paste, you would have to do a couple of extra steps to add the '.au' bit but that seems an unlikely scenario. It is easy enough to modify to suit this use case but I think immediate tokenization is better the majority of the time.

jackiehluo:
Hey! We've pushed an update (`0.4.47`) that should fix this issue—let us know if you see it again! Thanks!

PR #2635, opened 2016-07-18
Pasting into a participant field now preserves any text already there (Issue #2410)

jathpala:
This modifies the behavious when pasting into the to, cc or bcc fields of a new draft. This allows you to, for example, copy the domain of the email address from the clipboard and type the addressee and then paste the domain. Any text already in the input field is not automatically deleted. Rather the pasted
text is added to the already existing text. If the new contents of the email field now matches as an email address (based on RegExpUtils.emailRegex().test()), then the address is immediately tokenized. If the new input text doesn't match as an email, just append the new text and wait for further input.

### DIFF
--- a/src/components/tokenizing-text-field.cjsx
+++ b/src/components/tokenizing-text-field.cjsx
@@ -528,8 +528,13 @@ class TokenizingTextField extends React.Component
 
   _onPaste: (event) =>
     data = event.clipboardData.getData('text/plain')
-    @_addInputValue(data)
-    event.preventDefault()
+    newInputValue = @state.inputValue + data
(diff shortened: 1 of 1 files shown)


bengotow:
Hey! Thanks for contributing this. Long term, we're looking to clean email address specific logic out of TokenizingTextField (and move it to the subclass ParticipantsTextField) but it looks like there's already another reference to the email regex in `_onInputChanged`, so this doesn't make the situation any worse!

I believe there's a GitHub issue for this too, will see if I can find it and reference this PR. Cheers—

bengotow:
Marking https://github.com/nylas/N1/issues/2410 as done-pending-release!

grinich:
59ktq

PR #2659, opened 2016-07-22
Linux Mint 18 support.

pabloniklas:
Allows the deb package to be installed on Linux Mint 18. Xenial is not yet supported in nylas repos. Anyway, the postscript still works, creating the nylas.list files with the "vivid" version. I know Xenial is LTS (as Linux Mint 18 Sarah).


bengotow:
Looks good to me—thanks!

PR #2722, opened 2016-08-11
Adding nowrap to outline-view item names 

lukemwillis:
to prevent email address with hyphens from wrapping.

Currently:
Before

After adding nowrap:
After

### DIFF
--- a/static/components/outline-view.less
+++ b/static/components/outline-view.less
@@ -99,6 +99,7 @@
       overflow: hidden;
       line-height: @line-height-small;
       text-overflow: ellipsis;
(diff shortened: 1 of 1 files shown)


lukemwillis:
This does affect folder names in the menu as well.

Before:
Before

After:
After

grinich:
hi @lukemwillis -- we actually added this wrap on purpose so you can see the actual name of a label or folder. the left bar is pretty narrow so often the items are much longer than can be displayed

grinich:
as such we're unfortunately not going to merge this into mainline. but thanks for the contribution!

Issue #2726, opened 2016-08-12
Improve first name parser / don't show honorifics or name prefixes / Remove "mr, ms, dr, prof, etc"


chr-br:
When viewing a message, if the from/to field is unexpanded, what should show is the first names of all the senders/recipients. But if that person has a title (Mr., Dr., Prof., etc.) it is the title which shows up. 

Here are some images:

![title2](https://cloud.githubusercontent.com/assets/2973412/17634372/4ecfc8d0-60a0-11e6-9aa8-9135c294e871.png)
![title1](https://cloud.githubusercontent.com/assets/2973412/17634373/4ee6aad2-60a0-11e6-85a4-cae342e481f3.png)

As you can see, this functionality is not useful, since I can't actually see who's emailed me without expanding it.

CodeBytere:
@emorikawa @chr-br --> https://github.com/nylas/N1/pull/2732

PR #2732, opened 2016-08-17
improve first name parser by adding honorific exclusions

CodeBytere:
added exclusions for Dr., Mrs., Mr., Mx., Prof., and Ph.D. in the `firstName` method of `contact.coffee` so that their first name would not show up in emails as such


jackiehluo:
Oh, cool! Thanks for the PR—looks great!

emorikawa:
Closes #2726

PR #2741, opened 2016-08-19
Closes #2723 -> apm ignore plugin dev-dependencies

### LINKED ISSUE #2723: Plugin Failed to Install - Users reporting same error
##### Are there any related issues?

No related issues that I could find

OSX El Capitan

v0.4.47

**Bug?**
##### Do you have any third-party plugins installed? If so, which ones?

Nope, only the one that I can't install

N/A

##### Is the issue reproducible with a particular attachment, message, signature, etc?

Anytime you install https://github.com/colinking/n1-unsubscribe/releases/tag/1.5.5

I checked the logs and console and there is no output. There were errors initially, but I've fixed anything that popped up initially. Is there a time-out or other issue that is causing the install error? If I run `npm install` (or `apm install`) in my directory first, then install through the Nylas GUI, the plugin installs correctly, but without the pre-step, the Nylas GUI install fails When it fails to install, Nylas deactivates the package. After reloading and simply reactivating the package everything works.

Where is the apm install stdout logged to?

KyleKing:
Adds the `--production` flag to `apm install`. It should both be faster and allows people to install great plugins like, `n1-unsubscribe`!

*Currently, users have to install our plugin, reload Nylas, then go back to the plugin page to reactivate the plugin. The error is entirely due to all of the dev-dependencies currently installed by `apm install`, which shouldn't be downloaded anyway. See #2723 for how to reproduce the install problem*

I built and tested the fork locally:
nylas n1 electron today at 6 55 56 am

### DIFF
--- a/src/apm-wrapper.coffee
+++ b/src/apm-wrapper.coffee
@@ -219,7 +219,7 @@ class APMWrapper
         error.stderr = stderr
         callback(error)
 
-    apmProcess = @runCommand(['install'], {cwd: dir}, exit)
+    apmProcess = @runCommand(['install', '--production'], {cwd: dir}, exit)
     handleProcessErrors(apmProcess, errorMessage, callback)
 
   uninstall: (pack, callback) ->


KyleKing:
Hi, I just wanted to follow up. I hope this is a pretty easy fix and can get incorporated in the next release!

bengotow:
This looks great to me, and I've confirmed it also fixes the issue for me. Thanks for contributing this! Also, it might be worth submitting the same PR to Atom to see what they think there :-) This code in N1 is based on an older version of https://github.com/atom/settings-view/blob/06a022c84f870ae685b6ce594829a5af491414f3/lib/package-manager.coffee#L312

Issue #2884, opened 2016-10-01
Convert all Coffeescript to ES6


jackiehluo:
❌ to the ☕ script! ES6 is cleaner, more modern, and has a much more reliable linter—plus, people know how to read it—so we want to convert everything in the codebase. Most of the code is already there, but there are still specs written in Coffee.

krngrvr09:
Hello, I would like to work on this issue :)

zcserei:
I would also like to contribute!

krngrvr09:
Wow! There are 265 ``.coffee`` files. This will take some time I guess. But I agree its a good starter issue.

jonboiser:
Hi everyone, I just wanted to share what I did for the small PR I just submitted.

## Steps
1. Followed the [contribution guide](https://github.com/nylas/N1/blob/master/CONTRIBUTING.md) to get the app installed.
2. Ran the bootstrap script and test script.
3. In Atom, I was getting errors trying to use the local eslint setup, so had to install eslint, babel-eslint, and airbnb to my global npm. Not sure what is up with that.
4. Used [decaffeinate](https://github.com/decaffeinate/decaffeinate) to quickly convert the `coffee` file.
5. renamed that output to use the `es6` file extension
6. cleaned up the `es6` file
7. ran the `./N1.sh --test --enable-logging` commands a few times, changing the spec as a sanity check.

This particular spec was kind of weird, since it doesn't have an assertion, so I forced it to fail by adding `expect(true).toBe(false)` at the end as a sanity check.

jackiehluo:
Yay, thanks so much, all! Just FYI, particularly for these Coffeescript ➡️ ES6 conversions, you can also run the linter with `script/grunt lint`. We've had some trouble with `decaffeinate` in the past (since it's not always perfect), so it's good to do a sanity check by making sure the code is still working the same way.

BethyDiakabananas:
Is it too late to contribute to this? I would definitely like to contribute to this!

jackiehluo:
@BethyDiakabananas Not at all! You should!

oksas:
@jonboiser, thanks for the steps on conversion; it was super helpful in getting me started quickly. I am about to submit a PR for another small, single file conversion and thought it worth noting a few other things:

[...]

PR #2886, opened 2016-10-01
convert nylas-protocol-handler-spec to ES6

jonboiser:
for #2884 

I cherry picked the shortest spec to figure out a process for doing these conversions, which I'll document for others on the original issue.


bengotow:
Looks great! Thank you for doing this - we really appreciate CoffeeScript conversions. We're actually tracking our progress off of CoffeeScript here: https://docs.google.com/spreadsheets/d/1h-l3N5TjQXvH7VK37rdegirDFwgUqJwP49Ks_snUF0w/edit#gid=0. We originally omitted specs, but started including them recently.

PR #2914, opened 2016-10-09
bump(electron):Disable Sierra tabs

### LINKED ISSUE #2904: macOS Sierra "Tabs" feature does not work in N1
image

dcondrey:
Disables Sierra tab bars, resolves #2904

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/nylas/N1/issues"
   },
-  "electronVersion": "1.4.1",
+  "electronVersion": "1.4.2",
   "dependencies": {
     "async": "^0.9",
     "babel-core": "6.8.0",


grinich:
Hey @dcondrey -- bumping electron for us is a pretty complicated maneuver since we QA the entire app while doing it. 

We're definitely planning to keep upgrading Electron (as we've done in the past) but it doesn't make sense to make this change via a PR like this.

Better is probably to create a GitHub issue to bump electron (with your reason), and then you'll get notified when that is closed and the update goes out. 😄

grinich:
Also just to be clear: not disagreeing we should do this-- just need to do it a different way. :)

dcondrey:
Gotcha, had a feeling that would be the case.

bengotow:
Yep—thanks! We're definitely looking to do this, but also trying to see if we could support the tabs properly. Good to know that bumping Electron would do the trick! The next update (this week) will be a minor one, but we'll probably be doing a full QA pass for the release after that, and we could bump Electron then.

PR #2960, opened 2016-10-23
Make send action button title customizable 

koenpunt:
Currently the Send button always is "Send" or "Send + {ICON}". 

But when a ComposerExtension doesn't have an icon, or the icon is not descriptive enough for the action, there's no way of making clear what the send action is. 

Also, even though the documentation states: 

> It is used in the hover title text of your  option in the dropdown menu.

It is not, so maybe this title attribute should be added. Or this sentence should be removed.

### DIFF
--- a/internal_packages/composer/lib/send-action-button.jsx
+++ b/internal_packages/composer/lib/send-action-button.jsx
@@ -116,19 +116,19 @@ export default class SendActionButton extends React.Component {
     return {preferred, rest};
   }
 
-  _contentForAction = ({iconUrl}) => {
-    let plusHTML = "";
-    let additionalImg = false;
+  _contentForAction = ({buttonTitle, title, iconUrl}) => {
+    let additionalImg = null;
+
+    const actionTitle = buttonTitle ? buttonTitle : title;
 
     if (iconUrl) {
-      plusHTML = (<span>&nbsp;+&nbsp;</span>);
       additionalImg = (<RetinaImg url={iconUrl} mode={RetinaImg.Mode.ContentIsMask} />);
     }
 
     return (
       <span>
         <RetinaImg name="icon-composer-send.png" mode={RetinaImg.Mode.ContentIsMask} />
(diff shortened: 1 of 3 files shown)


internal_packages/send-and-archive/lib/send-and-archive-extension.cjsx
```diff
@@ -13,6 +13,7 @@ class SendAndArchiveExtension extends ComposerExtension
     if draft.threadId
       return {
         title: "Send and Archive"
+        buttonTitle: "Send +"
```
KevinMartin: Aren't you missing a space before the last double quote?

internal_packages/composer/lib/send-action-button.jsx
```diff
@@ -116,19 +116,24 @@ export default class SendActionButton extends React.Component {
...
+          &nbsp;
+          <RetinaImg url={iconUrl} mode={RetinaImg.Mode.ContentIsMask} />
+        </span>
+      );
     }
 
     return (
       <span>
         <RetinaImg name="icon-composer-send.png" mode={RetinaImg.Mode.ContentIsMask} />
-        <span className="text">Send{plusHTML}</span>{additionalImg}
```
KevinMartin: You removed "Send", but aren't setting it the default anywhere. If you don't have the Send+Archive option/extension enabled, will the button just be blank?

koenpunt: `actionTitle` will default to the `title` ([line 122](https://github.com/nylas/N1/pull/2960/files#diff-efe6eb7a1b18b8d286bd0e97c9a438e9R122)), which is mandatory.
The title was already defined for the `SendActionButton` component here: [internal_packages/composer/lib/send-action-button.jsx#L47](https://github.com/nylas/N1/blob/a5ff3cfc6216b0c5eb9fbbdec504992bb4cbb7c7/internal_packages/composer/lib/send-action-button.jsx#L47) but was not used for the button.

PR #2962, opened 2016-10-24
use single regex exec for toEquivalentEmailForm

koenpunt:
Because I think a single execution should be faster than 2 separate.


koenpunt:
I don't think the failed tests are related to my changes..

bengotow:
Hmm @emorikawa wrote this - maybe he can take a look. (Any reason we were using two regex's? From the blame it doesn't seem like it: https://github.com/nylas/N1/commit/b761afe96093e40c34621acff9d86e3be67c2dd9#diff-60a3c157cccc7cabf00265dcc7b997c2R214)

bengotow:
This looks great - thanks!

Issue #147, opened 2015-10-22
[Feature request] composition in plain text


mbolivar:
I'd like to use Nylas for reading and responding to patch emails on mailing lists. These often have strict requirements that message bodies be plain text. I don't see a way to set this up in Nylas.

grinich:
This currently isn't supported, as the Nylas Platform APIs only support HTML for sending. https://nylas.com/docs/platform/#sending

Note that a plain text version is included, but you can't compose in plain text.

mbolivar:
Thanks for getting back to me.

FWIW, this is a deal-breaker for me personally, but I wish you the best of luck nonetheless. Fighting the good fight on the desktop email front is something that nobody seems willing to do besides you guys these days.

screendriver:
+1

Same here. I need to send mails in plain text.

wget:
@grinich Do you have any solution used at Nylas to vote for feature requests like this one?
Like said in the aforementioned related bug I opened, having this feature is a deal-breaker/requirement on my side to adopt completely Nylas N1.

grinich:
Just to be clear, N1 _will_ send a `text/plain` MIME part for any message that goes out. We built the sync engine APIs to only support HTML for simplicity, but it "down converts" the HTML to a text version using the [html2text](https://github.com/Alir3z4/html2text/) Python package. You can see where we do it [here](https://github.com/nylas/sync-engine/blob/dae94ada8c7327255dabb952314a34e9bf13c2cc/inbox/sendmail/message.py#L93). 

We actually haven't found any issues with this and the conversion does pretty much what you would expect (e.g. turning blockquotes into `>` sections). For mail servers that don't use HTML, they can always grab the plain text section included in the raw MIME body.

Can you give us a bit more info on what this is stopping you from doing, or why it's a dealbreaker? Is the html2plain causing weird rendering somewhere? Is a mail server or email list rejecting your messages? If so, which one and what are they running? etc. etc. Thanks! :smile:

wget:
I know Nylas is sending the email in both version. This is what I directly checked when I tested this client for the first time ;-)

[...]

PR #2963, opened 2016-10-24
WIP: Plain text mail

koenpunt:
this is some groundwork for sending plain text emails from N1. I think
it now covers all the required headers for sending email, but I might
be wrong here.

I also added some basic specs, but these need some more work.

CC: #147


grinich:
Hi @koenpunt -- can you please re-open this PR once your tests pass? If you're looking for folks to help out, I recommend joining our [community Slack channel](http://slack-invite.nylas.com/).

PR #2996, opened 2016-10-30
Fix this line on non-POSIX systems, such as Windows

KevinMartin:
Per: https://github.com/nylas/N1/commit/f8f0cb206f53183698ee1648a7a789bfbf2598ac


bengotow:
Ahh thank you! Sorry about that, I thought it was available but a no-op.

PR #3020, opened 2016-11-07
(darkside) fixing composer fields spacing

jamiewilson:
- Removed styles for composer fields that were no longer useful
- reduced padding around 'attachment' icon in collapsed thread container
- updating attachment selector names for custom attachment hover styles


bengotow:
Awesome thank you!

PR #3024, opened 2016-11-08
Error handling for self hosting connect

Arekkusu-chan:
- If the xmlHttp status was not 200 then an error was not returned
- Added an else so if the connection to the server fails the ready state and http status will be displayed as an error
- This will help users debug issues when trying to set up the application and server for development


bengotow:
Hey! Thanks for contributing this - I'm not quite sure this logic is correct though. It's possible for an xmlHttp statusCode other than 200 to be returned from the server. For example, if you've misconfigured mysql and the sync engine is unable to write into the accounts table, it might return a 400 or 500. Mapping those type of scenarios to a "couldn't connect" message might make things more confusing!

Arekkusu-chan:
Hi,
Would you recommend a full set of error validation rules returning an individual statement for each type of error or should I just modify the sentence to 'An error was encountered on the server'?

I still feel any kind of error information will be less confusing than no error information at all since I spent a while debugging the code trying to find why my client could not connect to the sync engine and a simple 404 error being displayed would have saved me a lot of time.

Issue #3043, opened 2016-11-13
Compile error: Could not find pro submodule at ~/N1/script/src/pro. Skipping.**`


wilmcontreras:
##### What operating system are you using?
Ubuntu 16.04

##### What version of N1 are you using?
Latest (cloned from Git Nov 11th 2016)

Issue: Basically, when running ./bootstrap, I get the following output:

````
sudo ./bootstrap 

-- Running npm install in ./build with system config --
npm WARN nylas-build@ No license field.

-- Running npm install in ./apm with apm config --
npm WARN n1-bundled-apm@ No license field.

-- Running npm dedupe in ./apm/node_modules/npm with apm config --

-- Running npm install in ./. with electron config --
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: node-mac-notifier@0.0.13 (node_modules/node-mac-notifier):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for node-mac-notifier@0.0.13: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

-- Running npm dedupe in ./. with electron config --

**-- Linking Nylas submodule resources --
Could not find pro submodule at ~/N1/script/src/pro. Skipping.**`
````

I noticed that /src is in the parent directory, so, I made a copy in /script; however, /script/src/pro is empty anyways, so, not sure what bootstrap is supposed to find there.

jackiehluo:
That's expected behavior for the open-source project!

wilmcontreras:
I see... but nothing happens afterwards, I'm back in the Terminal. Where do I go from there?

wilmcontreras:
Got it sorted, had to run ` npm install -g express`

KevinMartin:
@jackiehluo This seems to be an issue in the builds, too...

https://ci.appveyor.com/project/nylas/n1/build/job/la0tiaqbmwv195pn

jackiehluo:
@KevinMartin I think the problem with that build is just the `node-mac-notifier` package, which we're working on fixing!

MooShoe:
I have the same issue, but @Khashir-E's solution isn't working. I'm on Ubuntu 16.04.

joepeak:
i am also,i'm on Windows 10 64

PR #3046, opened 2016-11-13
Windows x64

kevinmartin:
APM now supports x64. Atom is putting x64 in beta now.

https://github.com/atom/atom/pull/13135

### DIFF
--- a/apm/package.json
+++ b/apm/package.json
@@ -6,6 +6,6 @@
     "url": "https://github.com/nylas/n1"
   },
   "dependencies": {
-    "atom-package-manager": "1.15.0"
+    "atom-package-manager": "1.15.1"
   }
 }
--- a/apm/package.json
+++ b/apm/package.json
@@ -6,6 +6,6 @@
     "url": "https://github.com/nylas/n1"
   },
   "dependencies": {
-    "atom-package-manager": "1.1.1"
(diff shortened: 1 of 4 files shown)


KevinMartin:
There seems to be some sort of build issue with AppVeyor. Possibly related to #3043 and https://github.com/nylas/N1/commit/e89fd364b4706b4a0d9972a1c4ce2fa131eaa4c6

bengotow:
Hey! Thanks for submitting this - do you know of any major benefits to shipping 64-bit apps on Windows? Not questioning this effort or anything, I'm just honestly not sure. We've put off getting 64 bit support working because everyone seemed OK using the 32-bit version and N1 doesn't use enough memory to need the 64-bit address space.

KevinMartin:
Maybe a default install won't need the 64-bit address space, but when you add multiple plugins and need to search through hundreds of contacts/threads/emails/etc, there should be some benefit.

Atom team had a similar stance, but realized it was important, because most people don't use the default installation without plugins.

KevinMartin:
@bengotow the AppVeyor CI test is passing now.

Stanzilla:
Will this get merged?

KevinMartin:
Doubtful since the architecture for Nylas Mail has changed from Nylas N1/Pro

PR #3047, opened 2016-11-14
Rename Google Apps to G Suite in onboarding

jpelgrom:
Google renamed Google Apps (both 'normal' and education editions) to G Suite on 2016-09-29 ([announcement](https://blog.google/products/g-suite/all-together-now-introducing-g-suite/)). Updated text for account selection to match Google's branding.


bengotow:
Ahh man... as much as I hate "G Suite" as a product name, this should definitely happen. Thanks for the PR!

Issue #3100, opened 2016-11-28
Feature:  Notifications for items that are marked important (or inbox) only


yakshaving:
**Feature Request?**

Currently, users get notifications for every single spammy email list they might be a part of, or social media notifications.  Normally using Gmail, many of these emails would just go into the 'social' or 'promotions' inbox.   

It would be great if Apple notifications diverted from Nylas were only limited to items that are/were deemed "important" enough to make it to your inbox.  

People nowadays are overwhelmed by notifications.  This would be a great way to limit the number of distractions that people already have using a fairly simple heuristic.

Cybolic:
The current version already only shows notifications for emails that end up in the inbox. I've implemented the remaining part of the request, a config toggle to only show notifications for emails that arrive in the inbox and are marked important.

PR #3110, opened 2016-12-02
Implemented "only important" functionality and added checkbox to settings.

Cybolic:
Implements/fixes issue #3100 .


Cybolic:
The tests are currently failing because they're expecting to only see the previous configuration options. I've looked into fixing them, but they are a bit too hard-coded for what I have time for right now.

PR #3147, opened 2017-01-06
Fix link to Flux diagram in docs/Architecture.md

oksas:
Pretty minor change! It looks like the location of the Flux architecture diagrams got moved around on the Flux docs since the last changes on Architecture.md; this change updates the link here to the new/current location of the diagram.


grinich:
thx!

Issue #2486, opened 2016-06-19
Make it possible to apply mail rules to all inboxes 


devotox:
Make it possible to apply mail rules to all inboxes or a easy way to copy rules from one inbox to another

PR #3164, opened 2017-01-16
Adds option to reprocess *all* mail; not just inbox (with Mail Rules)

### LINKED ISSUE #3159: Add option to "Process entire mailbox", applying mail rules to archive
macOS 10.12.2; N1 0.4.402

**Bug?**
##### Do you have any third-party plugins installed? If so, which ones?
No

No

--

'Process entire mailbox' does not work in 'Mail Rules'.

It first appears with some small number, always the same, and then disappears - as if maybe it had finished (but far too quickly).

I cannot find any to which they've been applied though, other than new incoming emails.

OJFord:
And removes "entire" from the button that runs against inbox, which was the cause of my initial confusion, thinking it wasn't working.

Closes #3159 (& #2486?).

### DIFF
--- a/internal_packages/preferences/lib/tabs/preferences-mail-rules.jsx
+++ b/internal_packages/preferences/lib/tabs/preferences-mail-rules.jsx
@@ -67,7 +67,7 @@ class PreferencesMailRules extends React.Component {
     });
   }
 
-  _onReprocessRules = () => {
+  _onReprocessRules = (inboxOnly) => {
     const needsMessageBodies = () => {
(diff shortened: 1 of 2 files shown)


OJFord:
I don't know how well this will be styled - I'm having some trouble getting it to build (even on master) at the moment.

OJFord:
Closing because I can't test it currently, and I won't have any Mail Rules when I'm downgraded to freeloader edition in a few days anyway. Feel free to take over if you'd like this too!

PR #3175, opened 2017-01-17
Invert mail discussions order

mickaellecolazet:
Hello,

It would be great if it were possible to invert the order of viewing of the mail discussions.
Not from le oldest to the newest from top to bottom, but inverse.

Thanx.
Regards.


grinich:
hey @mickaellecolazet -- please vote up the existing GitHub issue rather than creating a PR for this feature request. thanks!

PR #3255, opened 2017-02-03
Update README.md to fix inconsistent references of N1

manicmarvin:
I've changed sentences that refer to N1 to refer ro Nylas Mail.

I'm unclear why and how there is no leading paragraph describing this renaming decision and preventing confusion. I would imagine part of re-branding would be to inform existing users of the re-brand (?) and that these things would be thought of before rolling out the re-brand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 ## Download Nylas Mail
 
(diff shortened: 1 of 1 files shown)


grinich:
Thanks (and sorry for the confusion). We're hoping to merge them together very shortly.We definitely thought about it a lot but figured it was better to get a free binary into the hands of our eagerly waiting users rather than wait for everything to be perfect.

PR #3349, opened 2017-04-06
Fix broken image url in README.md

binhonglee:
Seems like the website structure is changed entirely causing the url for the image in README.md to be broken.


binhonglee:
Can someone review this?
@grinich @jstejada

PR #3373, opened 2017-04-12
Add setting to disable offline notification

jacobmischka:
I don't believe Nylas Mail can be built from source yet, so I can't actually test this. But it looks right 🤷‍♂️ .

Obviously someone will need to test this, or please help me build from source.

Alleviates #3300


jacobmischka:
It doesn't look like travis failed because of anything I did, but I'm definitely not certain. Please let me know if there's anything that needs to be done.

PR #3388, opened 2017-04-14
Add flex resizing for html eiting window

thebearjew:
### Summary
Previously, when editing an email signature as HTML, the preview window
was extremely small (~ 168 x 90 px). The correct size should have been
that of its parent (.scroll-region-content-inner).

The textarea element created from the ContentEditable component is set
to `flex: 1`, which should result in it growing to its parent's size.

However because the immediate parent of the ContentEditable component
was not `display: flex`, the `flex-grow` attribute was not applied.

### Before (Screenshot)
image

### After (Screenshot)
image

---

**Note: Although I had intended to run the test-suite using `npm test` initialization process didn't work and I have not been able to resolve the issue.**


benkahle:
I think this is an exact duplicate of https://github.com/nylas/nylas-mail/pull/3238 :)

thebearjew:
@benkahle Ahaha, I didn't even notice! I'll close this one out 👍

Issue #3440, opened 2017-04-20
Launch on system start doesn't work on Windows


davidbgeek:
The _Launch on system start_ checkbox in Preferences creates a shortcut to `C:\Users\droiz\AppData\Local\nylas\Update.exe`, which is now an incorrect path. As well, changing the path to `C:\Users\droiz\AppData\Local\NylasMail\Update.exe` and keeping the arguments that Nylas includes in the path causes the _Get started_ page to open and a new profile of Nylas to be created, instead of just launching the existing instance of Nylas.

##### Are there any related issues?

N/A

##### What operating system are you using?
Windows 10 Build 16176

##### What version of Nylas Mail are you using?
2.0.16

**Bug?**
##### Do you have any third-party plugins installed? If so, which ones?
GitHub, Jiffy

##### Is the issue related to a specific email provider (Gmail, Exchange, etc.)?
N/A

##### Is the issue reproducible with a particular attachment, message, signature, etc?

N/A

lxalln:
Interestingly, I don't see that option in Preferences at all. Also using version 2.0.16 on Windows 10 (although 14393)
![image](https://cloud.githubusercontent.com/assets/2781908/25241940/b2006940-25ef-11e7-8a4e-1ab13ebfad7d.png)

davidbgeek:
@lxalln That's odd. Can you see anything between _Use Nylas Mail as default mail client_ and _Show icon in menu bar / system tray_ if you check Developer Tools?

lxalln:
Html:
```html
<section>
    <div class="item">
        <input type="checkbox" id="default-client" value="on">
        <label for="default-client">Use Nylas Mail as default mail client</label>
    </div>
    <!-- react-empty: 40489 -->
    <div class="item">
        <input type="checkbox" id="core.workspace.systemTray" value="on">
        <label for="core.workspace.systemTray">Show icon in menu bar / system tray</label>
    </div>
    ...
</section>
```

React is hiding the option, which leads me to:

[...]

PR #3443, opened 2017-04-20
Update _launcherPath to use correct folder & file

lxalln:
Looks like this got missed during a recent update.

Should fix: #3440

### DIFF
--- a/packages/client-app/src/system-start-service.es6
+++ b/packages/client-app/src/system-start-service.es6
@@ -104,7 +104,7 @@ class SystemStartServiceWin32 extends SystemStartServiceBase {
   }
 
   _launcherPath() {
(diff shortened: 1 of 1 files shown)


lxalln:
@jstejada do these get looked at? Seems you have a lot of people making issue and PR contributions, but no feedback at all, not even acknowledgement from Nylas.

dweremeichik:
@lxalln Was this still an issue on [Nylas Mail Lives](https://github.com/nylas-mail-lives/nylas-mail) I know your PR never made it in.

lxalln:
@dweremeichik As far as I'm aware this was still an issue. Although I never got around to running NML as a daily driver, so wasn't too concerned with it starting up on boot.

PR #3473, opened 2017-04-24
the email match should be case insensitive when selecting the alias

yehiasalam:
The _fromContactForReply selects the right email address to fill the from field when replying to messages. The emailMatch should be case insensitive when comparing the list of alias and the to field of the email. This solves the issue for ticket https://github.com/nylas/nylas-mail/issues/3462 and https://github.com/nylas/nylas-mail/issues/3454


dweremeichik:
@yehiasalam I know this PR is from some time ago. Nylas is no longer supporting this product, but we would love to have a PR for this against [Nylas Mail Lives](https://github.com/nylas-mail-lives/nylas-mail).

Issue #3155, opened 2017-01-13
You are not running Debian, Ubuntu, ElementaryOS or LinuxMint.  Not adding Nylas repository.


epocopa:
##### Are there any related issues?

#395
##### What operating system are you using?
Elementary OS Loki
##### What version of N1 are you using?
Last one available

**Bug?**
Im running Elementary OS Loki, and when I run: sudo dpkg -i N1.deb
It says: You are not running Debian, Ubuntu, ElementaryOS or LinuxMint.  Not adding Nylas repository.

phrz:
This issue persists as of today.

bmcclure:
And as of today as well. It'd be great to get this fixed in the production release of Nylas Mail.

dylanparry:
Does anyone know what the repository config is so we can at least add it manually? I didn't realise I wasn't getting any updates until now and have been stuck with an old version, which I've manually updated only today.

dylanparry:
To answer my own question, for Ubuntu at least, it's:

```
deb http://apt.nylas.com/ubuntu trusty main
```

The latest Ubuntu release I could find in the repository was _Trusty_. It appears to still work fine in later versions (I'm running 17.04, Zesty).

abstracted:
Same Issue
Elementary OS Loki

PR #3555, opened 2017-05-19
Fixes #3108 - Elementary Loki

### LINKED ISSUE #3108: Not adding Nylas repository - Elementary OS (Loki)
Upon install through the .deb, it says that I am not running ElementaryOS (I indeed am) and that it will not add the repo. Previously mentioned in #395.

> Setting up nylas (0.4.400) ...
You are not running Debian, Ubuntu, ElementaryOS or LinuxMint.  Not adding Nylas repository.

```
lsb_release -s -i
elementary
```

Based on #395, it looks like Elementary changed their format.

dweremeichik:
Elementary OS changed their release name when they released the "Loki" version of their OS. This change prevents Nylas from "installing" its repository for Loki. For versions previous to Loki, running `lsb_release -s -i` would return 

> elementary OS

In Loki, running `lsb_release -s -i` now returns

> elementary

This PR allows for both versions. Note I thought a simple "or" would be more efficient than a regex, and seems like a better choice than `"elementary"*`

See:
#3108
#3155

### DIFF
--- a/packages/client-app/build/resources/linux/debian/postinst
+++ b/packages/client-app/build/resources/linux/debian/postinst
@@ -24,7 +24,7 @@ case "$1" in
 
         DISTRO=`lsb_release -s -i`
 
-        if [ "$DISTRO" = "Ubuntu" ] || [ "$DISTRO" = "elementary OS" ] || [ "$DISTRO" = "LinuxMint" ] ; then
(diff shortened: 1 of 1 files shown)


dweremeichik:
Looks like you have an issue with openssl on your CI. @jstejada

dweremeichik:
Closed in favor of #3556.

Issue #3424, opened 2017-04-19
Option for custom port SMTP


therimalaya:
My institute uses Exchange server, but I can also use IMAP where I need to set the SMTP port to 26 and there is no option to set it in Nylas-mail. It could work that way. I could connect with exchange server too but it option is not availiable in the nylas-mail application in both linux and mac I am using.

##### Are there any related issues?

...

##### What operating system are you using?
...

##### What version of Nylas Mail are you using?
...

--

**Bug?**
##### Do you have any third-party plugins installed? If so, which ones?
...

##### Is the issue related to a specific email provider (Gmail, Exchange, etc.)?
...

##### Is the issue reproducible with a particular attachment, message, signature, etc?

...

--

**Feature Request?**
##### Does this feature exist in another mail client or tool you use?
...

Forsaken64:
I have almost the same problem, my mail server provider changed the default ports

jstejada:
hi! we will add this soon. thanks!

danielmelogpi:
Mine too... in fact, this is very common on enterprise on-premisses e-mail servers.

blabj:
Same here.  I need custom IMAP and SMTP ports!  IIRC this was doable in N1 when I tried it a while back.

jurem:
Same here

albarki:
same here

danielmelogpi:
Any news if this will be part of future releases/patches?

Jona-lee:
same here, I cannot connect my SMTP server without custom SMTP ports, please release paches soon.

emper0r:
Same here, we need to use custom port, put as variable, in that way the rest we can do to custom MTA ports even for receive, smtp/imap, thanks

Quintanar777:
same here

HugoReeves:
Is this coming?

voltarontas:
same here

vchicor0d:
same here

lasypig:
same here

axelvnk:
Opened up a pull request for this since I also wanted to add my own e-mail address with custom port.

bengotow:
Hey folks—Nylas Mail is no longer maintained, but this has been fixed in https://github.com/Foundry376/Mailspring/commit/e4d61462edf086e64a309a36321ef9bcfc8f464f on the [Mailspring](https://getmailspring.com) fork. You can connect to SMTP without auth and also on custom ports.

PR #3565, opened 2017-05-22
Enable free editing of the port field when configuring a general e-ma…

danielmelogpi:
This pull request enable users to freely edit the port they're connection to when configuring a custom e-mail account.
This is a farily common use case for enterprise e-mail accounts that are hosted by the company itself, as IT admins can change ports at will.

Solves issue #3424


dweremeichik:
@danielmelogpi Hey Dan, Happy Hacktoberfest! Would you consider opening a PR for this over at [Nylas Mail Lives](https://github.com/nylas-mail-lives/nylas-mail)?

Issue #3382, opened 2017-04-13
Error - Failed to save "config.json"


nirvaat:
Hi,

I have today installed the Desktop client on Windows 10. I have added 5 email accounts all from GSuite. It is working fine but I am getting an alert popup saying - "Failed to save Config.json". It is appearing randomly automatically. It gives two options "Okay" and "Try Again". 

I tried to look for the issue in Git and on Google, but nothing found related to this issue. Any help is appreciated. I am liking this application and would like to fix this irritating issue.

Thank you,
Rajendra

jstejada:
Thanks for reporting this! We'll look into it

nirvaat:
Thank you. Today there was an update in the app and I thought it is fixed but now there is another issue popping up. At random interval it is showing a red alert in bottom left saying "Unable to Authenticate with google for email."

emersonk:
I'm also having the first issue. Also installed on Windows 10, I have an imap account, a gmail account, and an outlook account linked. I can open another issue if that is preferred.

IsoPhoenix:
I also had this same problem. Pops up when I'm trying to change the account labels for my accounts, I try to type a new name but my text is immediately reverted to the original account label, kind of like lag. It took many many tries with clicking "Try Again" and eventually I could change the account names.

emersonk:
Nylas 2.0 update appears to have fixed this for me

jstejada:
Are people still seeing it on 2.0.15?

Thanks!

IsoPhoenix:
Yes I am still seeing it. Unable to save config.json

nirvaat:
Yes with new update also, the issue persists. When started it starts appearing at random intervals. When I checked there are lot of config files in the folder. And it creates a new file every few seconds, I am not sure why?

joeflateau:
+1

joeflateau:
FWIW, I renamed config.json to _config.json and the popup went away for a few minutes then came back

joeflateau:
Nevermind. I'm an idiot. I'm out of disk space.

IsoPhoenix:
I'm not out of disk space but the popup still occurs

dnlsrl:
I'm having the same error, which was not fixed with 2.0.18 update. I thought it was the custom theme I loaded into it, but I removed it and the problem persists.

I don't know if it's related or not, but my inbox seems to be in a permanent state of syncing.

There's also this error line in the detailed log, again, I don't know if related:

[...]

PR #3567, opened 2017-05-24
Replace single config.json with a buffer of files

seesemichaelj:
This makes it so that we never have to rename a config.json.bak file over the current config.json file. This also reduces the chance of file writing collusion. This is to prevent corrupt config.json files

This addresses issue #3382.

Instead of creating config.json.{randomId}.bak files and renaming them over config.json, this PR attempts to create a new config.json file (with a sortable timestamp). If there are more than 10 timestamped files, it will remove the oldest ones to maintain no more than 10 config files. The PR will successfully load old config.json files and adopt the new scheme (it leaves the old config.json files for safe keeping, but it never uses it again). When loading a config file, this PR will get the most recent config.timestamp.json file; if it fails, the user can now select "Try Older Configuration" to try the next recent file.

The format of the new filename is config.{millisecondsSinceEpoch*1000 + fileNumber}.json. The fileNumber is 0 by default and increments if the file already exists. Why I didn't use nanoseconds instead? I was thinking that a filenumber would eventually still be required and two saves back to back are more likely than separated saves (I kind of pulled this one out of the air, but it works).


nirvaat:
looks like there is no response from Nylas support. Can you provide the updated exe file for windows 10.

seesemichaelj:
turns out it's not as easy to build for Linux/Mac than Windows. Let me look into it and get back to you. I should respond by end of Friday with an update

nirvaat:
As it is a web application in a window, can't we change the JS code in the local already installed copy? If yes then what is the location of files?

seesemichaelj:
It's not necessarily a web app in a window; the Windows application is bundled up and the JS code is not modifiable. I looked in C:\Users\{USER}\AppData\Local\NylasMail folder where the exe is and could not find the file (i.e. config-persistence-manager.es6 and fs-utils.es6). Seems like they use Squirrel to build and maybe not as complicated as I thought.

seesemichaelj:
@nirvaat Please note that the code I'm working off of is 2.0.15 (2.0.32 is the latest version) because they haven't pushed their source code to Github in about a month

nirvaat:
did you able to get it working on windows?

seesemichaelj:
@nirvaat sorry I didn't get back to you. I was able to build the executable and create an installer, and the installer seemed to do something but I was unable to get it to load under Windows. It doesn't appear to really be initializing nylas mail when I run the executable, so I'm guessing I did something wrong in the build process. I've been busy so I haven't got a chance to look at it further. Hopefully the team will be able to merge this PR this week into an update since their update engine seem to be seamless. Your best bet is either to figure out how to build it yourself and try or push the team to get this reviewed and integrated until my schedule clears up (probably next week).

dweremeichik:
@seesemichaelj out of curiosity what was your build process? Do you have VS2013 installed?

seesemichaelj:
@dweremeichik I have VS2015 installed. The build process was fairly simple:

- `npm install`
- `npm build-client` (required running the git-bash/powershell/terminal as admin in windows for me)
  - this will build the application itself. from here you should be able to do `npm run client` to start the app. continue on for how to make the install exe. for linux/mac you should also have installer packages (deb/rpm for linux and dmg for mac)
- `node packages/client-app/build/create-signed-windows-installer.js` will create a signed installer for windows (at this point, i got a "file path/name is too long, windows sucks, we know it, blah blah blah" and had to move my repo to C:\ and run again. you should find an executable in the `packages/client-app/dist/` directory

when i tried to use/install the installer i was having issues getting the app to run. 3 nylas-mail processes would run after showing the install splash screen and then they'd idle and nothing would appear on the screen. I'm unsure about path forward for that. i figured it wouldnt take this long for the nylas team to review my commit and provide feedback and/or blast an update, but i guess they got busier things to do

dweremeichik:
@seesemichaelj wow, thanks for the the detailed writeup! More than what I was expecting. I think vs2015 might be your problem. Check out the appveyor.yml file notice `npm config set msvs_version 2013` I also remember reading in an older issue that the build would not work with vs2015. Worth a shot, and if you get it to work, we can host the exe here: https://github.com/nylas-mail-lives/nylas-mail

seesemichaelj:
@dweremeichik cheers, i'll give it a shot

seesemichaelj:
@dweremeichik I'm not sure if you would prefer your repo or the organization I set up a few weeks back. I have forked all of Nylas' repos at https://github.com/nylas-community

hktegner:
This is an absolute blocker if you haven't managed to sync messages once. Gives a very poor impression of the product. I've failed to setup even a single gmail account. Disappointing for something that looked to be really good.

seesemichaelj:
@hktegner I'm not sure if your comment was directed towards this Pull Request (i.e. me) or Nylas. The issue (#3382) is probably a better place to voice your concerns; but even more than that support@nylas.com.

If you want to follow the development on fixing this issue, you are better off following https://github.com/nylas-mail-lives/nylas-mail/issues/15 as a few of us are continuing development on Nylas over there.

